Issue #2, opened 2014-09-25
Consider removing process modifier from classes/filenames


shaun-edwards:
The current structures within Descartes are all prefixed with `process`.  However, the data types do not include any process data.

I'd like to propose that the `process` modifier be removed from the data types and file names and replaced with more generic types such as `waypoint` and `trajectory`.  Higher level classes will certainly contain process data, at which point, the prefix would seem more appropriate.

shaun-edwards:
@dpsolomon, please weigh in.  I'd like to make this change soon.

dpsolomon:
+1 I agree that process_\* is no longer appropriate

PR #5, opened 2014-09-30
Class package renaming

shaun-edwards:
Address issue #2.


shaun-edwards:
@dpsolomon, please review or let me know if I should assign to somebody else.

shaun-edwards:
@jrgnicho & @psturgeon, be aware of these name changes (they can be painful, if/when they get accepted).

psturgeon:
+1

Issue #3, opened 2014-09-25
Waypoints must support joint/cart types


shaun-edwards:
The current waypoint types are specific to Cartesian waypoints.  While the majority of a path is expected to be this type, joint waypoints, and other cooler waypoints we don't even know about, should be supported.  To support both joint and Cartesian waypoints, a common base class/interface is proposed.  Specific joint and Cartesian implementations will be created, while still allowing for new, custom types as well.

dpsolomon:
+1 I will take a look at this and maybe make a first pass. If someone else does it first, that would be fine as well.

dpsolomon:
This structure is not developed enough to either limit or promote this idea, but I have one option. The ProcessPt (or WayPoint as it may become in the near future) holds information about the different frames the robot cares about.  We could/should also add a vector of joint positions that satisfy the listed frames, and a function pointer to a "sampler" for that point.  The call to the sampler should populate and/or add to the list of allowable joint positions.

For a fixed joint point, the joint position vector can be fixed to size 1, and the sampler always returns immediately with the fixed point. Alternatively, the sampler does nothing and the joint position is fixed by some other method.

In either case, each WayPoint should be aware of its position in both Cartesian and Joint space at all times, because the use cases will always care about at least one, and most will care about both of these.

The Cartesian position is subtly important whether the point is a fixed joint position of not; When performing Cartesian (including circular) interpolation between points, it must be known what endpoint is being interpolated. Therefore, it is not enough to know that a point is fixed in joint space, but what tool point is being referenced for the path planning task.

Summary:
- Add variable for vector of joint positions that satisfy WayPoint.
- Add pointer to sampler.
- Add funtion to set fixed joint position.
- Add function to update frames based on fixed joint position.

dpsolomon:
Another thought as I submit the last comment:
The ProcessPt was meant to be independant of the robot that is executing the task. There should be another layer that asembles the graph of solutions, and that is probably the appropriate place for a fixed joint point. Perhaps the fixed joint point is a derived class of the ProcessPt?

shaun-edwards:
I found something in openrave that could be just what we are looking for (too bad the website is down).

[...]

PR #6, opened 2014-10-08
Issue#3 trajpt base class

dpsolomon:
There are still no methods for actually setting values, but this starts to build ideas and functionality we need. Fixes issue #3.


descartes_trajectory_planning/CMakeLists.txt
```diff
@@ -2,24 +2,31 @@ cmake_minimum_required(VERSION 2.8.3)
...
 
 
 catkin_package(
-#  INCLUDE_DIRS include
-#  LIBRARIES descartes_trajectory_planning
-#  CATKIN_DEPENDS roscpp
-#  DEPENDS system_lib
+  INCLUDE_DIRS
+    include
+  LIBRARIES
```
shaun-edwards, line 17: Is there much value in creating these as separate libraries?  Couldn't we just compile all the source files into a single library `descartes_trajectory_planning`?

descartes_trajectory_planning/include/descartes_trajectory_planning/cart_trajectory_pt.h
```diff
@@ -0,0 +1,171 @@
...
+  EIGEN_MAKE_ALIGNED_OPERATOR_NEW;
+public:
+  CartTrajectoryPt();
+  virtual ~CartTrajectoryPt();
+
+
+  /**@name Getters for Cartesian pose(s)
+   * @{
+   */
+
```
shaun-edwards, line 102: I'm not sure we need to document the implemented virtual methods.  Doxygen will do this by default ( http://www.stack.nl/~dimitri/doxygen/manual/config.html#cfg_inherit_docs ).  When using an interface definition (like we are), the behavior of the implemented method should match the interface documented behavior.

descartes_trajectory_planning/include/descartes_trajectory_planning/joint_trajectory_pt.h
```diff
@@ -0,0 +1,154 @@
...
+    return nominal-tol_below;
+  }
+
+  double range() const
+  {
+    return upperBound() - lowerBound();
+  }
+
+  double nominal;
+  double tol_above, tol_below;
```
shaun-edwards: Would prefer to keep the member variable nomenclature consistent with Cartesian (i.e. and upper/lower bound vs. tol_above/below).

descartes_trajectory_planning/include/descartes_trajectory_planning/trajectory_pt.h
```diff
@@ -25,91 +25,141 @@
 #ifndef TRAJECTORY_PT_H_
 #define TRAJECTORY_PT_H_
 
-#include <string>
-#include <moveit/kinematic_constraints/kinematic_constraint.h>
+//#include <string>
```
shaun-edwards: Remove commented string include ( if not used ).

descartes_trajectory_planning/include/descartes_trajectory_planning/trajectory_pt.h
```diff
@@ -25,91 +25,141 @@
...
 /**@brief A TrajectoryPt describes how a TOOL may interact with a PART to perform an automated trajectory.
  * The TOOL is something held by the robot. It is located relative to robot wrist/tool plate.
- * TOOL pose can change (e.g. robot holding workpiece) or be fixed (e.g. robot holding MIG torch).
- * Tool pose is described by fixed transform from wrist to tool_base, and variable transform from tool_base to tool_point.
  * The PART is something that exists in the world/global environment that is not held by robot.
- * PART is located relative to world coordinate system, and is described by
- * a (currently) fixed transform from world to part_base, and a variable transform from part base to specific point on part.
- * Variable transforms of both TOOL and PART have tolerances on both position and orientation.
- * Optionally, additional constraints can be placed on position and orientation that can limit, but not expand, existing tolerances.
+ * Particular descriptions of TOOL and PART are left to the derived classes.
```
shaun-edwards: Not sure I saw these descriptions.  `TOOL` and `PART` definitions for a joint point doesn't have much meaning that I can think of.  Also, a `WORLD` frame was introduced in the derived types, but not explained.

descartes_trajectory_planning/include/descartes_trajectory_planning/trajectory_pt.h
```diff
@@ -25,91 +25,141 @@
...
-  TrajectoryPt() {};
-  virtual ~TrajectoryPt() {};
-
-private:
-  Frame                         tool_base_;             // Fixed transform from wrist/tool_plate to tool base.
-  TolerancedFrame               tool_pt_;               // Underconstrained transform from tool_base to effective pt on tool.
-  Frame                         object_base_;           // Fixed transform from WCS to base of object.
-  TolerancedFrame               object_pt_;             // Underconstrained transform from object base to goal point on object.
-  TrajectoryTransitionPtr          transition_;            // Velocities at, and interpolation method to reach this point
+  TrajectoryPt();
```
shaun-edwards: If we never expect to create a base Trajectory pt (i.e. this is the base/interface class), then constructors/destructors are not needed and should be removed.

descartes_trajectory_planning/include/descartes_trajectory_planning/trajectory_pt.h
```diff
@@ -25,91 +25,141 @@
...
+   * @{
+   */
+
+  /**@brief Get single Cartesian pose associated with closest position of this point to seed_state.
+   * (Pose of TOOL point expressed in WORLD frame).
+   * @param pose If successful, affine pose of this state.
+   * @param seed_state RobotState used for kinematic calculations and joint_position seed.
+   * @return True if calculation successful. pose untouched if return false.
+   */
+  virtual bool getClosestCartPose(Eigen::Affine3d &pose, const moveit::core::RobotState &seed_state) const = 0;
```
shaun-edwards, line 84: Is there a use case for this method?

descartes_trajectory_planning/include/descartes_trajectory_planning/trajectory_pt.h
```diff
@@ -25,91 +25,141 @@
...
+  /**@name Getters for joint pose(s)
+   * @{
+   */
+
+  /**@brief Get single Joint pose closest to seed_state.
+   * @param joint_pose Solution (if function successful).
+   * @param seed_state RobotState used for kinematic calculations and joint position seed.
+   * @return True if calculation successful. joint_pose untouched if return is false.
+   */
+  virtual bool getClosestJointPose(std::vector<double> &joint_pose, const moveit::core::RobotState &seed_state) const = 0;
```
shaun-edwards, line 110: Use case?

descartes_trajectory_planning/include/descartes_trajectory_planning/trajectory_pt.h
```diff
@@ -25,91 +25,141 @@
...
+   * @param id Number to set id_ to.
+   */
+  void setID(size_t id)
+  {
+    id_ = id;
+  }
+  /** @} (end section) */
+
+protected:
+  size_t                        id_;                    /**<@brief ID associated with this pt. Generally refers back to a process path defined elsewhere. */
```
shaun-edwards, line 155: I am considering specifying a unique ID, since an ordered ID would have to be shifted when points were added/removed and therefore not unique.  Probably a discussion for a later PR.

descartes_trajectory_planning/include/descartes_trajectory_planning/trajectory_pt.h
```diff
@@ -25,91 +25,141 @@
...
+  TrajectoryPtTransitionPtr     transition_;            /**<@brief Velocities at, and interpolation method to reach this point **/
+
+  /** @name TrajectoryPt transforms. Note that base class does not include any method to set these members, as derived classes may have specific implementations.
+   *  @{
+   */
+  Frame                         tool_;                  /**<@brief Transform from robot wrist to active tool pt. */
+  Frame                         wobj_;                  /**<@brief Transform from world to active workobject pt. */
```
shaun-edwards: Inconsistent naming: `work object`, `world`, `part`.

descartes_trajectory_planning/src/trajectory_pt.cpp
```diff
@@ -27,4 +27,14 @@
 namespace descartes
```
shaun-edwards: May not need source file for this at all (if we still with it as an interface class and simple methods defined in the header).

shaun-edwards:
Minor issues and mostly just requests for clarifications.  Thanks for the contribution.

descartes_trajectory_planning/include/descartes_trajectory_planning/trajectory_pt.h
dpsolomon: They are required if we ever want to do linear interpolation. Also, I expect that getNominalCartPose() will want to return a pose similar to CartTrajPt, so you will need to have the same frame information.

descartes_trajectory_planning/include/descartes_trajectory_planning/trajectory_pt.h
dpsolomon, line 110: An A\* style graph search would probably want something like this. At the least, it is req'd for Cartesian pts, but I think it would be used for joint points too.

descartes_trajectory_planning/CMakeLists.txt
dpsolomon, line 17: Interesting, I hadn't thought of that. Will do.

descartes_trajectory_planning/include/descartes_trajectory_planning/cart_trajectory_pt.h
```diff
@@ -0,0 +1,171 @@
...
+  {}
+  double x_upper, y_upper, z_upper, x_lower, y_lower, z_lower;
+
+  void clear() {x_upper = y_upper = z_upper = x_lower = y_lower = z_lower = 0.;};
+};
+
+/**@brief TolerancedFrame extends frame to include tolerances and constraints on position and orientation.
+ * Samplers that are called on this object should sample within tolerance, and check if result satisfies constraints.
+ */
+struct TolerancedFrame: public Frame
```
ClayFlannigan, line 67: I haven't been following too closely, so pardon the interruption, question: are we planning to handle process information (gun on, close gripper etc.) with each trajectory point?  Perhaps a ProcessFrame?

dpsolomon, line 67: My feeling is that process data lives outside of this framework because it isn't required for planning. Each trajectory point contains an id_ member. Presumably, the actual application maps each of its own process points to a trajectory pt, and activates process information when the apropriate trajectory point is reached.

dpsolomon:
I have the structural changes done, I will complete the aesthetic changes later today.

jrgnicho:
I see that there are several methods that take a moveit::core::RobotState object in order to perform actions specific to a particular robot model.  It is my understanding that this RobotState object holds one or more "planning groups" and most of its IK and FK related methods require knowing about the group http://moveit.ros.org/doxygen/classmoveit_1_1core_1_1RobotState.html .  Thus, our current interface methods won't be of much use as they are since "planning groups" can not be specified with them.  One easy solution is to expand each method's argument list so that they accept planning group names or moveit::core::JointModelGroups objects.  Another option is to pass the RobotState and planning group name in the constructor of any specialized class that inherits from these interfaces and remove the RobotState object from the methods' argument list since it would be redundant to have it there.  The second option is more appealing to me because it would make our interfaces more flexible by not being tied up to any particular robot model implementation (the moveit RobotState as of now). One good reason for keeping our interface classes agnostic is those cases where multiple joint configurations exist for a single end-effector pose.  The moveit RobotState only returns one solution (at least I haven't found a way to get more than one) even if more are available.  Ideally, we'd handle this by implementing a custom ik solver or robot model and save a reference to it in the specialized TrajectoryPt implementation.  This probably requires further discussion and should be addressed in future PRs.

dpsolomon:
This reminds me of why I wanted to separate these methods from the
TrajectoryPt class and write separate solvers to implement them.  However,
we have started down this path and I think it deserves bearing out.

We have talked about an inherited class of robot state that would address
most of these problems. Whatever we do, we should try to keep the
particular robot separate from the TrajectoryPt as much as possible.

I agree that most of this is appropriate for a separate PR.

descartes_trajectory_planning/include/descartes_trajectory_planning/cart_trajectory_pt.h
shaun-edwards, line 67: I hadn't considered using the `id_` member, but this is a good idea.  The approach I was considering was to create new waypoint classes that inherit from these base classes.  The new classes could still be utilized by the planners but also include process specific data.

gavanderhoorn, line 67: I agree with @dpsolomon, I think process data should be separated from trajectory data as much as possible. In general I believe working as abstract as possible will be advantageous.

dpsolomon:
Ready for re-review to resolve the current issue #3 . We should probably attack the modified RobotState class (i.e. Jorge's comment) shortly.

shaun-edwards:
+1...I'll add an issue for the RobotState.  I have an idea for how to handle this...you should see a PR from me soon.

PR #7, opened 2014-10-10
Initial addition of boost graph structures and creation.

psturgeon:


descartes_trajectory_planning/include/descartes_trajectory_planning/planning_graph.h
```diff
@@ -26,49 +26,114 @@
...
+  // TODO: 'overloaded' requests depending on source and destination
+  //bool GetShortestPathJointToCartesian(int startIndex, int endIndex, double &cost, std::vector<TrajectoryPt> &path);
+  //bool GetShortestPathCartesianToCartesian(int startIndex, int endIndex, double &cost, std::vector<TrajectoryPt> &path);
+
+  /**@brief Utility function for printing the graph to the console
+   * NOTE: should add other formats for output
+   */
+  void printGraph();
+
+private:
```
shaun-edwards: We typically used `protected` instead of `private` unless we are absolutely sure we don't want a sub-class to have access to the method.  This doesn't matter much at this point, but it may in the future.

descartes_trajectory_planning/include/descartes_trajectory_planning/planning_graph.h
```diff
@@ -26,49 +26,114 @@
...
+  moveit::core::RobotState *robot_state_;
```
shaun-edwards, line 109: We should use a smart pointer, instead of a raw pointer in most cases.  See http://www.boost.org/doc/libs/1_56_0/libs/smart_ptr/shared_ptr.htm  

Typically we `typedef` them.  See https://github.com/ros-industrial-consortium/descartes/blob/hydro-devel/descartes_trajectory_planning/include/descartes_trajectory_planning/cart_trajectory_pt.h#L31

descartes_trajectory_planning/include/descartes_trajectory_planning/planning_graph.h
```diff
@@ -26,49 +26,114 @@
...
 public:
+	// TODO: add constructor that takes RobotState as param
   PlanningGraph() {};
   virtual ~PlanningGraph() {};
+
+  /** @brief initial population of graph trajectory elements
+   * @param points list of trajectory points to be used to construct the graph
+   * @param cartesianPointLinks for each TrajectoryPt in the points list, a struct that indicates order of the points
+   */
+  bool insertGraph(std::vector<CartTrajectoryPt> *points, std::vector<CartesianPointRelationship> *cartesianPointLinks);
```
shaun-edwards: Could we simply assume the points are in order (removing the need for the second parameter)?

descartes_trajectory_planning/src/planning_graph.cpp
```diff
@@ -24,7 +24,279 @@
...
+//
+//	}
+
+// TODO: optionally take these as params to the constructor
+	bool PlanningGraph::insertGraph(std::vector<CartTrajectoryPt> *points, std::vector<CartesianPointRelationship> *cartesian_point_links)
+	{
+		// validate input
+		if(!points || !cartesian_point_links)
+		{
+			// one or both are null
```
shaun-edwards: In all failure cases, we should log an appropriate error.  See [here](https://github.com/ros-industrial-consortium/descartes/blob/hydro-devel/descartes_trajectory_planning/src/joint_trajectory_pt.cpp#L96) for a logging example.

<EDIT> I'm not sure if the logging wrappers allow for stream-like logging.  We might not use them if this is the case (probably just worth an issue at this point).

descartes_trajectory_planning/include/descartes_trajectory_planning/planning_graph.h
```diff
@@ -26,49 +26,114 @@
...
+  // TODO: removeTrajectory
+
+  /** @brief Calculate and return the shortest path from the given joint solution indices
+   * @param startIndex The index of the joint solution at which to start
+   * @param endIndex The index of the joint solution at which to end
+   * @param cost The cost of the returned path
+   * @param path The sequence of points (joint solutions) for the path (TODO: change to JointTrajectoryPt?)
+   * @return True if a valid path is found
+   */
+  bool getShortestPathJointToJoint(int startIndex, int endIndex, double &cost, std::list<int> &path);
```
shaun-edwards: One helpful overload would simply be from beginning to end of the Cartesian path.

descartes_trajectory_planning/include/descartes_trajectory_planning/planning_graph.h
psturgeon: Yes, the graph class can maintain the link order if the original list can be assumed to be in order.

descartes_trajectory_planning/src/planning_graph.cpp
```diff
@@ -24,7 +24,279 @@
...
+
+			// Starting from the destination point step through the predecessor map
+			// until the source point is reached.
+			while (current != start_index)
+			{
+				current = predecessors[current];
+				path.push_front(current);
+			}
+		}
+		return true;
```
shaun-edwards: Can this function fail?  What happens if the `if` statement above evaluates to false?

shaun-edwards:
Addition items:
- [ ] Reformat using eclipse http://wiki.ros.org/IDEs#Auto_Formatting
- [ ] Clarify index (I think this means path order) and id (which is unique and independent of order).  These seem to be used interchangeably, but I'm not sure they are.
- [ ] Use smart pointers (i.e. boost pointers) where possible instead of raw pointers.

I like the simple public interface.  I think we should stick with a minimal interface until we know better. 

Do we have any sense of the time it will take to solve a worst case graph.  I'm thinking 2000 cartesian points X 10 joint solutions/cartesian point = 20,000 vertices.

shaun-edwards:
@dpsolomon, do you have time to weigh in?

descartes_trajectory_planning/src/planning_graph.cpp
psturgeon: Addressed:  
else
  {
    logError("unable to find a valid path from start(%d) to end(%d)", start_index, end_index);
    return false;
  }

descartes_trajectory_planning/src/planning_graph.cpp
psturgeon: Logs added for each failure (that was already being handled).

descartes_trajectory_planning/include/descartes_trajectory_planning/planning_graph.h
psturgeon: Addressed. Changed private to protected.

descartes_trajectory_planning/include/descartes_trajectory_planning/planning_graph.h
dpsolomon: The relationship variable might make it easier to add/remove points from the graph. I don't know if we actually have a use case for this though...

dpsolomon: I think we can make this function take a vector of TrajectoryPt, as I believe everything it calls is generic to our base/derived class structure.

descartes_trajectory_planning/src/planning_graph.cpp
```diff
@@ -24,7 +24,279 @@
...
+		}
+		for(std::vector<CartesianPointRelationship>::iterator link_iter = cartesian_point_links->begin(); link_iter != cartesian_point_links->end(); link_iter++)
+		{
+			cartesian_point_link_[link_iter->id] = *link_iter;
+		}
+
+		// after populating maps above (presumably from cartesian trajectory points), calculate (or query) for all joint trajectories
+		if(!calculateJointSolutions())
+		{
+			// failed to get joint trajectories
```
dpsolomon: Unless I am misunderstanding something here, I believe the term "joint trajectories" used here and in following code should be replaced by "joint solutions" to avoid confusion. calculateJointSolutions() calculates the different joint poses that satisfy each TrajectoryPt.

dpsolomon:
WRT stream-like error logs, you can always create a stream object, push a bunch of stuff into it, and then export it to a string that is printed in the error. It is slightly longer than directly streaming an error, but it keeps us dependant on console_bridge rather than roscpp, which I think is a worthwhile tradeoff.

descartes_trajectory_planning/src/planning_graph.cpp
```diff
@@ -24,7 +24,279 @@
...
+
+			std::list<int> start_joint_ids = trajectory_point_to_joint_solutions_map_[start_cart_id];
+			std::list<int> end_joint_ids = trajectory_point_to_joint_solutions_map_[end_cart_id];
+
+			for(std::list<int>::iterator start_joint_iter = start_joint_ids.begin(); start_joint_iter != start_joint_ids.end(); start_joint_iter++)
+			{
+				for(std::list<int>::iterator end_joint_iter = end_joint_ids.begin(); end_joint_iter != end_joint_ids.end(); end_joint_iter++)
+				{
+					double transition_cost;
+					// TODO: Make a call to somewhere that takes to JointTrajectoryPts (std::vector<double>) to get a single weight value back
```
dpsolomon: Note: I think this weight function should be passed by function pointer to calculateEdgeWeights(), or set in the graph class. Different applications will use a different weighting function. We should come up with a prototype that matches the use cases we see now. This issue can wait until a later PR to resolve.

dpsolomon:
+1 I think this is what we need to get started with the graph. I think this can be merged as is, but I wouldn't complain if we wanted to wait until it was genericized to TrajectoryPt instead of CartTrajectoryPt.

dpsolomon:
Another thought: I think we can have the edge weights calculated by dijkstra during the solve step. This would allow us to pass the cost function with our just prior to the solve step without modifying the underlying graph. I think this is an example of an external graph property. Any thoughts?

ClayFlannigan:
One use case to consider is editing just one node. If you calculate cost at solve time, would this require recalculation of all the costs?  Or maybe they would be "cached"?

dpsolomon:
Good point. Also, if we stored them internally it would be easier to do a *
search with incremental graph generation. And we could always have a method
that recalculated all the costs.

Very well, internal cost storage it is!

shaun-edwards:
Some data on graph solving (this was done on a Windows box with the boost library in debug mode).  The absolute numbers are probably the right order of magnitude.  The relative comparison is probably more telling.  From this data it's clear we probably can't solve a dense graph (I'm pretty sure we all knew this).  We will have to solve a sparse graph with some method of "interpolating" to waypoints in between.

Interesting though, is that the time to populate the graph is as long as it is.  The graph population was not calling for IK or collision detection.  I believe it assigned random joint values.  It could clearly get much worse.

time_test_01

shaun-edwards:
+1... to keep things going.

shaun-edwards:
@psturgeon, please delete your branch if you are done (or you can continue to use it for PRs).

Issue #9, opened 2014-10-18
Parameter ordering


shaun-edwards:
The function/method parameter ordering does not conform to the ROS style [guide](http://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Function_Parameter_Ordering).  According to the guide, the parameters must be inputs first, outputs last.  Inputs should be passed as const references/points and outputs as non-const.

Affected files: 
- [x] trajectory_pt.h
- [x] cart_trajectory_pt.h
- [x] joint_trajectory_pt.h

jrgnicho:
Good point on the argument arrangement.  One thing to keep in mind is that passing the RobotState object as a const argument limits the methods that can be called since not all of them have the "const" keyword at the end of each method signature.  If we end up implementing our own Robot Model interface, we should make sure that the "const" attribute is adequately applied where needed.

Issue #8, opened 2014-10-14
moveit::core::RobotState dependency


jrgnicho:
This is the issue related to the discussion started in PR #6; in order to avoid repeating myself below I'm pasting the comment I made regarding this issue:

> I see that there are several methods that take a moveit::core::RobotState object in order to perform actions   specific to a particular robot model. It is my understanding that this RobotState object holds one or more "planning groups" and most of its IK and FK related methods require knowing about the group  http://moveit.ros.org/doxygen/classmoveit_1_1core_1_1RobotState.html . Thus, our current interface methods won't be of much use as they are since "planning groups" can not be specified with them. One easy solution is to expand each method's argument list so that they accept planning group names or moveit::core::JointModelGroups objects. Another option is to pass the RobotState and planning group name in the constructor of any specialized class that inherits from these interfaces and remove the RobotState object from the methods' argument list since it would be redundant to have it there. The second option is more appealing to me because it would make our interfaces more flexible by not being tied up to any particular robot model implementation (the moveit RobotState as of now). One good reason for keeping our interface classes agnostic is those cases where multiple joint configurations exist for a single end-effector pose. The moveit RobotState only returns one solution (at least I haven't found a way to get more than one) even if more are available. Ideally, we'd handle this by implementing a custom ik solver or robot model and save a reference to it in the specialized TrajectoryPt implementation. This probably requires further discussion and should be addressed in future PRs.

Another suggested solution by @shaun-edwards is to implement a RobotModel interface class that will be used as input for the TrajectoryPt interface methods instead of the moveit::core::RobotState as it is the case now.  This interface could then be specialized to handle the unique requirements of a path planner , such as returning  multiple joint configurations for a tool pose or it could even contain an instance of the moveit RobotState and it's corresponding planning group.  Hopefully, I captured the original idea correctly (@shaun-edwards  correct me if I'm wrong).  I'm on board with this approach
Something else that stood out to me were the constraint members used by the ToleranceFrame structures.  These constraints are also part of the moveit library and required a moveit::core::RobotModel instance (http://moveit.ros.org/doxygen/classmoveit_1_1core_1_1RobotModel.html). If we decide to use our own RobotModel interface then we might have to define our own constraint structures which do not depend on moveit's RobotModel class

PR #11, opened 2014-10-18
Added robot model interface class

shaun-edwards:
Addresses issues #8 and #9.

### DIFF
--- a/descartes_trajectory_planning/include/descartes_trajectory_planning/cart_trajectory_pt.h
+++ b/descartes_trajectory_planning/include/descartes_trajectory_planning/cart_trajectory_pt.h
@@ -97,29 +97,41 @@ public:
 
 
   /**@name Getters for Cartesian pose(s)
-   * @{
-   */
+     * @{
+     */
 
-  virtual bool getClosestCartPose(Eigen::Affine3d &pose, const moveit::core::RobotState &seed_state) const;
+    //TODO complete
+    virtual bool getClosestCartPose(const std::vector<double> &seed_state,
+                                      const RobotModel &model, Eigen::Affine3d &pose) const;
 
-  virtual bool getNominalCartPose(Eigen::Affine3d &pose, const moveit::core::RobotState &seed_state) const;
+    //TODO complete
+    virtual bool getNominalCartPose(const std::vector<double> &seed_state,
+                                      const RobotModel &model, Eigen::Affine3d &pose) const;
 
-  virtual void getCartesianPoses(EigenSTL::vector_Affine3d &poses, const moveit::core::RobotState &state) const;
-  /** @} (end section) */
+    //TODO complete
+    virtual void getCartesianPoses(const RobotModel &model, EigenSTL::vector_Affine3d &poses) const;
+    /** @} (end section) */
 
-  /**@name Getters for joint pose(s)
-   * @{
-   */
+    /**@name Getters for joint pose(s)
+     * @{
+     */
 
-  virtual bool getClosestJointPose(std::vector<double> &joint_pose, const moveit::core::RobotState &seed_state) const;
+    //TODO complete
+    virtual bool getClosestJointPose(const std::vector<double> &seed_state,
+                                       const RobotModel &model,
+                                       std::vector<double> &joint_pose) const;
+    //TODO complete
+    virtual bool getNominalJointPose(const std::vector<double> &seed_state,
+                                       const RobotModel &model,
+                                       std::vector<double> &joint_pose) const;
 
-  virtual bool getNominalJointPose(std::vector<double> &joint_pose, const moveit::core::RobotState &seed_state) const;
+    //TODO complete
+    virtual void getJointPoses(const RobotModel &model,
+                                 std::vector<std::vector<double> > &joint_poses) const;
+    /** @} (end section) */
 
-  virtual void getJointPoses(std::vector<std::vector<double> > &joint_poses, const moveit::core::RobotState &state) const;
-  /** @} (end section) */
-
-  virtual bool isValid(const moveit::core::RobotState &state) const;
-
+    //TODO complete
+    virtual bool isValid(const RobotModel &model) const;
   /**@brief Set discretization. Cartesian points can have position and angular discretization.
    * @param discretization Vector of discretization values. Must be length 2 or 6 (position/orientation or separate xyzrpy).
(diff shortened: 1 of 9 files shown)


shaun-edwards:
@jrgnicho, please review.

descartes_trajectory_planning/include/descartes_trajectory_planning/cart_trajectory_pt.h
```diff
@@ -97,29 +97,41 @@ class CartTrajectoryPt : public TrajectoryPt
...
+    //TODO complete
+    virtual bool getClosestCartPose(const std::vector<double> &seed_state,
+                                      const RobotModel &model, Eigen::Affine3d &pose) const;
+
+    //TODO complete
+    virtual bool getNominalCartPose(const std::vector<double> &seed_state,
+                                      const RobotModel &model, Eigen::Affine3d &pose) const;
+
+    //TODO complete
+    virtual void getCartesianPoses(const RobotModel &model, EigenSTL::vector_Affine3d &poses) const;
```
jrgnicho, line 112: How is the getCartesianPoses method going to be used? In the case of the Joint Trajectory point it makes intuitive sense to ask for multiple joint configurations for the same pose but it's not quite clear to me how you can ask for multiple poses?

shaun-edwards, line 112: The cart/joint point can have associated tolerances.  By sampling within those tolerance zones multiple Cartesian poses can be returned.

descartes_trajectory_planning/include/descartes_trajectory_planning/robot_model.h
```diff
@@ -0,0 +1,96 @@
...
+
+  /**
+   * @brief Returns "all" the joint poses("distributed" in joint space) for a desired affine pose.
+   * "All" is determined by each implementation (In the worst case, this means at least getIK).
+   * "Distributed" is determined by each implementation.
+   * @param pose Affine pose of TOOL in WOBJ frame
+   * @param joint_pose Solution (if function successful).
+   * @return True if successful
+   */
+  virtual bool getAllIK(const Eigen::Affine3d &pose, std::vector<double> &joint_pose) const;
```
jrgnicho: Are all joint values for each solution going to be appended to the joint_pose array? For instance, a 6DOF robot containing 4 valid joint configurations will return a joint_pose array of size 24?

shaun-edwards: Good catch...it should be a vector of vectors (like in the `trajectory_pt` methods)

descartes_trajectory_planning/include/descartes_trajectory_planning/robot_model.h
```diff
@@ -0,0 +1,96 @@
...
+   * @param pose Affine pose of TOOL in WOBJ frame
+   * @return True if valid
+   */
+  virtual bool isValid(const Eigen::Affine3d &pose);
+
+};
+
+}//descartes
+
+#endif /* ROBOT_KINEMATICS_H_ */
```
jrgnicho, line 96: Any reason why some of these interface methods are purely virtual (  = 0 ) while others aren't?

shaun-edwards, line 96: That all should be purely virtual (my mistake).

descartes_trajectory_planning/include/descartes_trajectory_planning/robot_model.h
```diff
@@ -0,0 +1,96 @@
...
+   * @return True if valid
+   */
+  virtual bool isValid(const std::vector<double> &joint_pose);
+
+  /**
+   * @brief Performs all necessary checks to determine affine pose is valid
+   * @param pose Affine pose of TOOL in WOBJ frame
+   * @return True if valid
+   */
+  virtual bool isValid(const Eigen::Affine3d &pose);
```
jrgnicho: I'm gonna take a leap of faith and assume that these "isValid" methods will be used in conjunction with the "isValid" methods of the Trajectory Point classes; if those methods take a "const" robot model argument then shouldn't the "isValid" methods in the RobotModel have the "const" attribute as well.  Not doing so will prevent us from using those methods.

shaun-edwards: Yep...making them const.

descartes_trajectory_planning/include/descartes_trajectory_planning/trajectory_pt.h
```diff
@@ -103,28 +107,37 @@ class TrajectoryPt
...
 
-  /**@brief Check if state satisfies trajectory point requirements. */
-  virtual bool isValid(const moveit::core::RobotState &state) const = 0;
+  /**@brief Check if state satisfies trajectory point requirements.
+   * @param model Robot model  object used to determine validity
+   */
+  virtual bool isValid(const RobotModel &model) const = 0;
 
   /**@brief Set discretization. Note: derived classes interpret and use discretization differently.
    * @param discretization Vector of discretization values.
```
jrgnicho, line 143: This is something that probably should've been discussed much earlier. but the type std::size_t used for the "id_" members will only allows us to assign positive values starting at zero.  This could be somewhat limiting if we had to check for ID validity by comparing its value to a range of valid id values.  If "ints" were used instead, we could default this ID value to -1 or less that zero which indicates that the point has not yet been indexed or initialized properly.

shaun-edwards, line 143: The `id_` should not be confused with index.  The ID simply must be unique and will be set upon construction (it will always be valid).  I don't think we will need to track the index internal to the trajectory point.  The trajectory itself will keep the order.  The `id_` will be used to look up points by _key_, so it must be unique.

descartes_trajectory_planning/include/descartes_trajectory_planning/cart_trajectory_pt.h
jrgnicho, line 112: Got it. 
+1

descartes_trajectory_planning/include/descartes_trajectory_planning/trajectory_pt.h
jrgnicho, line 143: +1

jrgnicho:
Looks good to me
+1

Issue #4, opened 2014-09-25
Position/orientation nominal values


shaun-edwards:
The  process point/waypoint structure doesn't capture nominal values.  A waypoint will likely have a desired value, with some error bound or tolerance.  When at all possible, the nominal value should be held.  The tolerance zone should only be used when the process requires/allows for it.

shaun-edwards:
@dpsolomon, please comment.

dpsolomon:
-1 each waypoint contains a set of frames and toleranced frames (inherits from frames). A frame is mostly just an affine - that is the default pose if the point.

I will write up the thought I had behind this layout in the readme, I think it will be clearer how it might work then.

PR #13, opened 2014-10-20
Implement derived trajpts

dpsolomon:
Not all methods are fully implemented, but it should be a start. Should also mostly address issue #4.

### DIFF
--- a/descartes_trajectory_planning/include/descartes_trajectory_planning/trajectory_pt.h
+++ b/descartes_trajectory_planning/include/descartes_trajectory_planning/trajectory_pt.h
@@ -44,6 +44,8 @@ struct Frame
   Frame(){};
   Frame(const Eigen::Affine3d &a):
     frame(a), frame_inv(a.inverse()) {};
+  Frame(const Frame &a):
+    frame(a.frame), frame_inv(a.frame_inv) {};
 
   EIGEN_MAKE_ALIGNED_OPERATOR_NEW;
   Eigen::Affine3d frame;
@@ -64,6 +66,7 @@ struct Frame
  * and use these process points to create a Trajectory that can be solved for a robot path.
  * In order to implement this easily, each process point should keep track of the TrajectoryPt id, and
  * provide an interpolation method between points.
+ *
  */
 class TrajectoryPt
 {
@@ -72,11 +75,11 @@ public:
   virtual ~TrajectoryPt() {};
 
   /**@name Getters for Cartesian pose(s)
+   * References to "closest" position are decided by norm of joint-space distance.
    * @{
    */
 
   /**@brief Get single Cartesian pose associated with closest position of this point to seed_state.
-   * (Pose of TOOL point expressed in WOBJ frame).
    * @param pose If successful, affine pose of this state.
    * @param seed_state RobotState used for kinematic calculations and joint_position seed.
    * @return True if calculation successful. pose untouched if return false.
@@ -84,21 +87,15 @@ public:
   virtual bool getClosestCartPose(Eigen::Affine3d &pose, const moveit::core::RobotState &seed_state) const = 0;
 
   /**@brief Get single Cartesian pose associated with nominal of this point.
-    * (Pose of TOOL point expressed in WOBJ frame).
     * @param pose If successful, affine pose of this state.
     * @param seed_state RobotState used for kinematic calculations and joint_position seed.
     * @return True if calculation successful. pose untouched if return false.
     */
   virtual bool getNominalCartPose(Eigen::Affine3d &pose, const moveit::core::RobotState &seed_state) const = 0;
-
-  /**@brief Get "all" Cartesian poses that satisfy this point. Use RobotState for performing kinematic calculations.
-   * @param poses Note: Number of poses returned may be subject to discretization used.
-   * @param state RobotState used for kinematic calculations.
-   */
-  virtual void getCartesianPoses(EigenSTL::vector_Affine3d &poses, const moveit::core::RobotState &state) const = 0;
   /** @} (end section) */
 
   /**@name Getters for joint pose(s)
+   * References to "closest" position are decided by norm of joint-space distance.
    * @{
    */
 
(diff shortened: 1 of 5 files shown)


descartes_trajectory_planning/CMakeLists.txt
```diff
@@ -40,7 +40,6 @@ add_library(descartes_trajectory
             src/cart_trajectory_pt.cpp
             src/joint_trajectory_pt.cpp
             src/trajectory.cpp
-            src/planning_graph.cpp
```
shaun-edwards, line 43: Did you mean to take this out?  Is it failing to build?

descartes_trajectory_planning/include/descartes_trajectory_planning/trajectory_pt.h
```diff
@@ -28,8 +28,8 @@
 #include <Eigen/Core>
 #include <Eigen/Geometry>
 #include <eigen_stl_containers/eigen_stl_vector_container.h>
+#include <moveit/robot_state/robot_state.h>
 #include <vector>
-#include "descartes_trajectory_planning/robot_model.h"
```
shaun-edwards, line 32: Did you mean to do this as well?  Perhaps you need to rebase/pull?

descartes_trajectory_planning/src/cart_trajectory_pt.cpp
```diff
@@ -38,57 +38,44 @@ CartTrajectoryPt::CartTrajectoryPt():
...
 }
 
-bool CartTrajectoryPt::getClosestJointPose(const std::vector<double> &seed_state,
-                                           const RobotModel &model,
-                                           std::vector<double> &joint_pose) const
+bool CartTrajectoryPt::getNominalJointPose(std::vector<double> &joint_pose, const moveit::core::RobotState &seed_state) const
 {
-  NOT_IMPLEMENTED_ERR;
-  return false;
+  moveit::core::RobotState state(seed_state);
```
shaun-edwards, line 60: This whole section will be replaced by a descartes `RobotModel` type (see PR #11).  However, this know tells me how use robot state to get what I want.

descartes_trajectory_planning/src/joint_trajectory_pt.cpp
```diff
@@ -36,56 +36,57 @@ JointTrajectoryPt::JointTrajectoryPt():
...
-  NOT_IMPLEMENTED_ERR;
-  return false;
-}
-
-bool JointTrajectoryPt::getNominalJointPose(const std::vector<double> &seed_state,
-                                            const RobotModel &model,
-                                            std::vector<double> &joint_pose) const
-{
-  NOT_IMPLEMENTED_ERR;
+  moveit::core::RobotState state(seed_state);
```
shaun-edwards, line 56: See comment above.

shaun-edwards:
@dpsolomon, this certainly helps me understand how to get the desired information out of moveit.  There  was a PR (#11) that collides with some of these changes.  I'm in the middle of porting the code that you added to a Moveit wrapper class for Descartes RobotModel.

descartes_trajectory_planning/CMakeLists.txt
dpsolomon, line 43: No, just a late-night mistake in merging.

descartes_trajectory_planning/include/descartes_trajectory_planning/trajectory_pt.h
dpsolomon, line 32: (same as above)

descartes_trajectory_planning/src/cart_trajectory_pt.cpp
dpsolomon, line 60: Good, that was the purpose of filling this in at all, as there is not enough information to actually do what we want with the RobotState as is. For one, we need to know what group we are working with, either by a string name of JointModelGroup

dpsolomon:
OK. Should I wait for that and merge with this, or fix the CMakeLists mistakes and let you merge your stuff after this?

shaun-edwards:
I'll create a new PR using what you started.  Should be done tomorrow.

dpsolomon:
Merge is fixed in this PR. Whichever way you want to proceed...

descartes_trajectory_planning/src/joint_trajectory_pt.cpp
```diff
@@ -96,38 +97,6 @@ bool JointTrajectoryPt::isValid(const RobotModel &model) const
     }
   }
   return true;
-  */
-  NOT_IMPLEMENTED_ERR;
-  return false;
-}
-
-bool JointTrajectoryPt::setDiscretization(const std::vector<double> &discretization)
```
shaun-edwards, line 104: Any reason you took this out?

shaun-edwards:
This PR is covered partially by PR #14.  A second PR will implement methods for joint and Cartesian points.

descartes_trajectory_planning/src/cart_trajectory_pt.cpp
```diff
@@ -38,57 +38,44 @@ CartTrajectoryPt::CartTrajectoryPt():
...
 }
 
-bool CartTrajectoryPt::getNominalCartPose(const std::vector<double> &seed_state,
-                                          const RobotModel &model, Eigen::Affine3d &pose) const
+bool CartTrajectoryPt::getNominalCartPose(Eigen::Affine3d &pose, const moveit::core::RobotState &seed_state) const
 {
-  NOT_IMPLEMENTED_ERR;
-  return false;
+  /* Simply return wobj_pt expressed in world */
+  pose = wobj_base_.frame * wobj_pt_.frame;
```
shaun-edwards, line 49: Should this include the tool inverses?  Does this cover the "puzzle" piece example?

shaun-edwards:
I've manually ported these to new branch.  Just the remaining questions about removing the "discretization" and the "nominal cartesian" methods.

shaun-edwards:
@dpsolomon, ping...see comment above.

descartes_trajectory_planning/src/cart_trajectory_pt.cpp
dpsolomon, line 49: At this point, the only use case I see for getNominalCartPose() method is finding 6DOF distance between adjacent trajectory points. This would be useful for, say, deciding how the path needed to be discretized, or setting a rough velocity.  This is not the call to make in order to do arm IK.

dpsolomon:
@shaun-edwards  Sorry, I thought this PR was mostly defunct. Se my note on the nominal Cartesian pose. I removed all get*Poses() and discretization references because those items are solver properties, not trajectory.  The trajectory points should be held as independant from the samplers and solvers as possible. They should really only have basic methods such as querying the nominal pose, and getting/setting various internal properties.  The sampler/solver should use the information from the TrajectoryPts to determine the robot solution.

shaun-edwards:
I didn't see that the get*Poses() were removed.  They are still required.

I agree that `trajectory_pt` should be independent of `solvers` and `samplers`, but I'm not sure it can work this way.  The main rationale for making a `trajectory_pt` look like both a cartesian and joint point is to allow planners to operate on trajectories of mixed types.  In order for this to work, a point needs to be able to calculate it's own IK/FK (and sample for that matter).

dpsolomon:
@shaun-edwards Ah, right, now I remember what I wanted to do. I wanted to wait until your Moveit adaptor was pulled first though.

Each `TrajectoryPt` should contain a pointer to a `sampler` class, and implement a `bool sample(vector<JointPoint>&, int n)` method that simply calls the sampler with that point.  The `sampler` should be able to incrementally sample according to its own discretization (or maybe discretization could also be passed as an arguement?). Repeated calls to `sample(.., 1) will return single-item vector<JointPoint>, sample(.., 0) will return a vector of all joint solutions according to the discretization.

The initial implementation should just call sample(.., 1) on each point, and the samplers should return the nominal joint solution. My though was that subsequent calls to sample() return results farther from the nominal until the tolerance is all used up.

You said you were working on a new branch. Do you want to merge that and then attack this, or do it all at once? (My preference is to merge what is there first). If so, we should close this PR and open yours for comment.

shaun-edwards:
@dpsolomon, I've opened a PR #15.  It does not address this discussion.  I will add an issue so we can keep moving forward.  I think we are in agreement in how to move forward though.  Just a few thoughts, but I'll save that for the issue.

PR #14, opened 2014-10-21
Added Moveit adapter package

shaun-edwards:
This implements the Descartes `RobotModel` interface for MoveIt.  The PR includes some parts of PR #13, but not all.  Method implementations for `TrajectoryPt` are not included.  However, these methods will simply be wrappers around RobotModel calls when required.

### DIFF
--- a/descartes_moveit/CMakeLists.txt
+++ b/descartes_moveit/CMakeLists.txt
@@ -5,6 +5,7 @@ find_package(catkin REQUIRED COMPONENTS
   console_bridge
   moveit_core
   tf
+  cmake_modules
   descartes_trajectory_planning
 )
 
--- a/descartes_moveit/CMakeLists.txt
+++ b/descartes_moveit/CMakeLists.txt
@@ -2,13 +2,12 @@ cmake_minimum_required(VERSION 2.8.3)
 project(descartes_moveit)
 
 find_package(catkin REQUIRED COMPONENTS
-  console_bridge
   moveit_core
   tf
   cmake_modules
   descartes_trajectory_planning
 )
-
+find_package(console_bridge REQUIRED)
 find_package(Boost REQUIRED)
 find_package(Eigen REQUIRED)
 
@@ -20,9 +19,12 @@ catkin_package(
     descartes_moveit
   CATKIN_DEPENDS
     moveit_core
+    tf
+    cmake_modules
+    descartes_trajectory_planning
   DEPENDS
-    Boost
     console_bridge
+    Boost
     Eigen
 )
 
--- a/descartes_moveit/CMakeLists.txt
+++ b/descartes_moveit/CMakeLists.txt
@@ -1,0 +1,53 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(descartes_moveit)
+
+find_package(catkin REQUIRED COMPONENTS
+  console_bridge
+  moveit_core
+  tf
+  descartes_trajectory_planning
+)
+
+find_package(Boost REQUIRED)
+find_package(Eigen REQUIRED)
+
+
+catkin_package(
+  INCLUDE_DIRS
+    include
+  LIBRARIES
+    descartes_moveit
+  CATKIN_DEPENDS
+    moveit_core
+  DEPENDS
+    Boost
+    console_bridge
+    Eigen
+)
+
+###########
+## Build ##
+###########
+include_directories(include
+                    ${catkin_INCLUDE_DIRS}
+                    ${Boost_INCLUDE_DIRS}
+                    ${Eigen_INCLUDE_DIRS}
+)
+
+## Descartes MoveIt lib
+add_library(descartes_moveit
+            src/moveit_state_adapter.cpp
+)
+target_link_libraries(descartes_moveit
+                      ${catkin_LIBRARIES}
+)
+
+
+#############
+## Testing ##
+#############
+
+# catkin_add_gtest(${PROJECT_NAME}-test test/test_descartes_trajectory_planning.cpp)
+# if(TARGET ${PROJECT_NAME}-test)
+#   target_link_libraries(${PROJECT_NAME}-test ${PROJECT_NAME})
+# endif()
--- a/descartes_moveit/include/descartes_moveit/moveit_state_adapter.h
+++ b/descartes_moveit/include/descartes_moveit/moveit_state_adapter.h
@@ -1,0 +1,110 @@
+/*
+ * Software License Agreement (Apache License)
+ *
+ * Copyright (c) 2014, Dan Solomon
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef MOVEIT_STATE_ADPATER_H_
+#define MOVEIT_STATE_ADPATER_H_
+
+#include "descartes_trajectory_planning/robot_model.h"
+#include "moveit/robot_model/robot_model.h"
+#include "moveit/kinematics_base/kinematics_base.h"
+#include <string>
+
+namespace descartes_moveit
+{
+
+/**@brief MoveitStateAdapter adapts the MoveIt RobotState to the Descartes RobotModel interface
+ *
+ */
+class MoveitStateAdapter : public descartes::RobotModel
+{
+public:
+
+  /**
+   * Constructor for Moveit state adapters (implements Descartes robot model interface)
+   * @param robot_state robot state object utilized for kinematic/dynamic state checking
+   * @param group_name planning group name
+   * @param tool_frame tool frame name
+   * @param wobj_frame work object frame name
+   */
+  MoveitStateAdapter(const moveit::core::RobotState & robot_state, const std::string & group_name,
+                    const std::string & tool_frame, const std::string & wobj_frame);
+  virtual ~MoveitStateAdapter()
+  {
+  }
+  ;
+
+  virtual bool getIK(const Eigen::Affine3d &pose, const std::vector<double> &seed_state,
+                     std::vector<double> &joint_pose) const;
+
+  virtual bool getAllIK(const Eigen::Affine3d &pose, std::vector<std::vector<double> > &joint_poses) const;
+
+  virtual bool getFK(const std::vector<double> &joint_pose, Eigen::Affine3d &pose) const;
+
+  virtual bool isValid(const std::vector<double> &joint_pose) const;
+
+  virtual bool isValid(const Eigen::Affine3d &pose) const;
+
+protected:
+
+  /**
+   * @brief Default constructor hidden
+   */
+  MoveitStateAdapter()
+  {
+  }
+  ;
+
+  /**
+   * Gets IK solution (assumes robot state is pre-seeded)
+   * @param pose Affine pose of TOOL in WOBJ frame
+   * @param joint_pose Solution (if function successful).
+   * @return
+   */
+  bool getIK(const Eigen::Affine3d &pose, std::vector<double> &joint_pose) const;
+
+  /**
+   * @brief Pointer to moveit robot state (mutable object state is reset with
+   * each function call
+   */
+  mutable moveit::core::RobotStatePtr robot_state_;
+
+  /**
+   * @brief Planning group name
+   */
+  std::string group_name_;
+
+  /**
+   * @brief Tool frame name
+   */
+  std::string tool_base_;
+
+  /**
+   * @brief Work object/reference frame name
+   */
+  std::string wobj_base_;
+
+  /**
+   * @brief Joint solution sample iterations for returning "all" joints
+   */
+  size_t sample_iterations_;
+
+};
+
+} //descartes_moveit
+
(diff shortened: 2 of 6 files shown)


descartes_moveit/include/descartes_moveit/moveit_state_adapter.h
```diff
@@ -0,0 +1,110 @@
...
+   * @param joint_pose Solution (if function successful).
+   * @return
+   */
+  bool getIK(const Eigen::Affine3d &pose, std::vector<double> &joint_pose) const;
+
+  /**
+   * @brief Pointer to moveit robot state (mutable object state is reset with
+   * each function call
+   */
+  mutable moveit::core::RobotStatePtr robot_state_;
```
shaun-edwards, line 85: I made this `mutable` to get around the const methods issues (`robot_state_` non-const methods are required to calculate things like IK).  I'm not sure of all the implications of this.  One potential issue (maybe only in C++11) is that this class is not thread-safe (see [here](http://channel9.msdn.com/posts/C-and-Beyond-2012-Herb-Sutter-You-dont-know-blank-and-blank), about 25:00 minutes in).  This should be captured in an issue or if anybody has any suggestions, I'm open.

descartes_moveit/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(descartes_moveit)
+
+find_package(catkin REQUIRED COMPONENTS
+  console_bridge
+  moveit_core
+  tf
+  descartes_trajectory_planning
```
gavanderhoorn, line 8: You're missing a dependency on `cmake_modules`: `FindEigen.cmake` is not part of CMake itself, but is distributed as part of `cmake_modules`, a ROS package.

descartes_moveit/package.xml
```diff
@@ -0,0 +1,60 @@
...
+  <!-- Use buildtool_depend for build tool packages: -->
+  <!--   <buildtool_depend>catkin</buildtool_depend> -->
+  <!-- Use run_depend for packages you need at runtime: -->
+  <!--   <run_depend>message_runtime</run_depend> -->
+  <!-- Use test_depend for packages you need only for testing: -->
+  <!--   <test_depend>gtest</test_depend> -->
+  <buildtool_depend>catkin</buildtool_depend>
+  <build_depend>console_bridge</build_depend>
+  <build_depend>moveit_core</build_depend>
+  <build_depend>tf</build_depend>
```
gavanderhoorn, line 46: Missing dep on `cmake_modules` here as well.

descartes_moveit/src/moveit_state_adapter.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+  if (robot_model_->hasJointModel(group_name_))
+  {
+    std::vector<std::string> joint_names = robot_model_->getJointModelGroup(group_name_)->getJointModelNames();
+    if (tool_base_ != joint_names.back())
+    {
+      logError("Tool: %s does not match group tool: %s, functionality will be implemented in the future",
+               tool_base_.c_str(), joint_names.front().c_str());
+    }
+    if (wobj_base_ != joint_names.back())
```
jrgnicho, line 48: Is the wobj_base not a link outside the planning group?

descartes_moveit/src/moveit_state_adapter.cpp
```diff
@@ -0,0 +1,180 @@
...
+  bool rtn = false;
+  if (robot_state_->setFromIK(robot_state_->getJointModelGroup(group_name_), pose, tool_base_))
+  {
+    joint_pose.resize(robot_state_->getVariableCount());
+    for (size_t ii = 0; ii < robot_state_->getVariableCount(); ++ii)
+    {
+      joint_pose[ii] = robot_state_->getVariablePosition(ii);
+    }
+    rtn = true;
+  }
```
jrgnicho, line 75: This way of retrieving the joint values feels uncertain since it is not associated with a particular planning group.  It's very likely that the robot state "variables" and "joints" are identical sets most of the time but It's presumable that there are some cases where this doesn't hold.  I'd recommend using the "copyJointGroupPositions(const std::string& joint_group_name, std::vector<double>& gstate)" [here](http://moveit.ros.org/doxygen/classmoveit_1_1core_1_1RobotState.html#aab69ae00f6d9c0d0249299115afcf7a0) or one of its variants since this returns the joints for the specified planning group.

descartes_moveit/src/moveit_state_adapter.cpp
```diff
@@ -0,0 +1,180 @@
...
+  joint_poses.clear();
+  for (size_t sample_iter = 0; sample_iter < sample_iterations_; ++sample_iter)
+  {
+    robot_state_->setToRandomPositions();
+    std::vector<double> joint_pose;
+    if (getIK(pose, joint_pose))
+    {
+      std::vector<std::vector<double> >::iterator it;
+
+      it = find(joint_poses.begin(), joint_poses.end(), joint_pose);
```
jrgnicho: I presume that this search is carried out by using the overloaded std::vector equals operator [here](http://www.cplusplus.com/reference/vector/vector/operators/)  in order to determine equality.  This may lead to multiple solutions that are effectively the same but still fail the overloaded equality comparison due to negligible differences in floating point precision.  An alternative would be to compare the difference against a small threshold in order to determine if they are close enough to be considered equal.

descartes_moveit/src/moveit_state_adapter.cpp
```diff
@@ -0,0 +1,180 @@
...
+  else
+  {
+    return true;
+  }
+}
+
+bool MoveitStateAdapter::getFK(const std::vector<double> &joint_pose, Eigen::Affine3d &pose) const
+{
+  bool rtn = false;
+  robot_state_->setVariablePositions(joint_pose);
```
jrgnicho: the setJointGroupPositions(..) [here](http://moveit.ros.org/doxygen/classmoveit_1_1core_1_1RobotState.html#aaf00ec8b048e7b23638858d2c1e0f56a) method should be used here in order to apply this to the planning group.

descartes_moveit/src/moveit_state_adapter.cpp
```diff
@@ -0,0 +1,180 @@
...
+  }
+  return rtn;
+}
+
+bool MoveitStateAdapter::isValid(const std::vector<double> &joint_pose) const
+{
+  bool rtn = false;
+  if (robot_state_->getVariableCount() == joint_pose.size())
+  {
+    robot_state_->setVariablePositions(joint_pose);
```
jrgnicho: See comment above

descartes_moveit/src/moveit_state_adapter.cpp
shaun-edwards, line 48: For this first implementation, work object base and tool base must be the start & end of the kinematic chain.  This will change in the future.

jrgnicho, line 48: In that case, should wobj_base_ be checked against joint_names.front() ?

descartes_moveit/src/moveit_state_adapter.cpp
dpsolomon: Agreed. Depending on the solver (ie KDL) all solutions would pretty much be unique if checked this way.

descartes_moveit/src/moveit_state_adapter.cpp
dpsolomon: This should work fine as well...

descartes_moveit/src/moveit_state_adapter.cpp
dpsolomon, line 48: I would rather see wobj_base_ and tool_base_ always be just outside the 1st and last link (respectively). They can be initialized to Identity by default.

descartes_moveit/src/moveit_state_adapter.cpp
```diff
@@ -0,0 +1,180 @@
...
+  }
+  else
+  {
+    logError("Could not set Cartesian pose.");
+    rtn = false;
+  }
+  return rtn;
+}
+
+bool MoveitStateAdapter::getAllIK(const Eigen::Affine3d &pose, std::vector<std::vector<double> > &joint_poses) const
```
dpsolomon, line 84: I have a thought, and it may be induced by the delirium of slogging through boost function pointers and binding all day, but here goes (this is also inspired by something I saw in moveit, but I will have to dig it up some other time). Create a defaultAllIkSolver(Affine3d, vector<double>) implementation that contains this code. Then change this signature to getAllIK(Affine3d, vector<double>, boost::function<void (Affine3d, double)> = &defaultAllIkSolver). 
This would allow us to pass in a public function of an IKFast library as an alternative to random sampling. By default (passing no new function) this call would operate the same as here.

descartes_moveit/src/moveit_state_adapter.cpp
shaun-edwards, line 48: @dpsolomon, in this instance `wobj_base_` and `tool_base_` are frame id's (strings).  I think they will be needed in case where the ik solver base/tip links do not match those assumed by trajectory points.

For example, our use case has a calibrated tool frame and world frame, defined in the URDF.  This means the trajectory point will have be relative to these frames.  This will result in a cartesian point with wobj and tool initialized to identity.

descartes_moveit/src/moveit_state_adapter.cpp
shaun-edwards, line 84: I'll add an issue for this.  My approach is bad for several reasons.  I'm just trying to get something that will quickly create multiple solutions to populate the graph.

descartes_moveit/CMakeLists.txt
shaun-edwards, line 8: I added this dependency only to find out that I didn't have `cmake_modules` installed.  Everything worked before adding the dependency, so I'm not sure it is really needed.  I think the dependency on `moveit_core` pulls in `Eigen`, though I'm not sure.

gavanderhoorn, line 8: > I added this dependency only to find out that I didn't have `cmake_modules` installed. 

catkin itself contains a CMake config file (`eigen-config.cmake`) for Eigen in Hydro (see [here](https://github.com/ros/catkin/tree/groovy-devel/cmake/Modules), `groovy-devel` is also used for Hydro). This was changed in Indigo, where a `FindEigen.cmake` was added to `cmake_modules`, and `catkin` no longer provides the config file for Eigen.

So not depending on `cmake_modules` works in Groovy & Hydro, but will stop working in Indigo (as there is no `FindEigen.cmake` anywhere anymore otherwise). Adding the dependency on `cmake_modules` will not affect Hydro compatibility, but will make your package more future proof.

See the Indigo [migration](http://wiki.ros.org/indigo/Migration#cmake_modules) guide, and the `cmake_modules` [readme](https://github.com/ros/cmake_modules).

>  Everything worked before adding the dependency, so I'm not sure it is really needed. I think the dependency on `moveit_core` pulls in `Eigen`, though I'm not sure. 

I've seen Dirk Thomas tell people multiple times that they should not depend on transitive dependencies, because they can change and are not under your control.

I agree that the Eigen dependency is ubiquitous, and unlikely to change soon, but in principle I agree with Dirk. Be explicit about your dependencies.

descartes_moveit/package.xml
shaun-edwards, line 46: Added.

descartes_moveit/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+find_package(Eigen REQUIRED)
+
+
+catkin_package(
+  INCLUDE_DIRS
+    include
+  LIBRARIES
+    descartes_moveit
+  CATKIN_DEPENDS
+    moveit_core
```
gavanderhoorn, line 21: I think you're missing `tf`, ~~`console_bridge`~~, `cmake_modules` and `descartes_trajectory_planning` here.

No: `console_bridge` is not actually a catkin package. Are you allowed to list it in the `find_package(catkin ..)` call then?

descartes_moveit/src/moveit_state_adapter.cpp
dpsolomon, line 48: The design of the tool/wobj base was to allow planning with either a pure
robot moveit package, or a custom one. In order to support that strategy, I
think what was missing is a frame_id for each of the base frames.

I would rather see the frame id's added to the adaptor, and an identity
frame used for the bases. That should help avoid future problems expanding
the usage.

The calls to actually get ik should always solve between two known frames.
Moveit::robotstate does some sort of automatic check and correction for
difference between the planning frames and ik solver frames

descartes_moveit/CMakeLists.txt
shaun-edwards, line 21: Corrected

shaun-edwards:
@jrgnicho, please review.  I've added issue #17 to address the IK searching.

descartes_moveit/src/moveit_state_adapter.cpp
shaun-edwards, line 84: See issue #17

shaun-edwards:
When `setFromIK` is called with default values (i.e. 0), the function uses the group defaults for `attempts` and `timeout` are used.  See [here](http://moveit.ros.org/doxygen/robot__state_8cpp_source.html#l01278).

descartes_moveit/src/moveit_state_adapter.cpp
jrgnicho, line 48: The current branch hydro-devel shows that the wobj_base_ is checked against the last element of the link names array "back()" .  According to the discussion above, should it be compared against the robot base instead or the first element in the link names array "front()"?

shaun-edwards, line 48: @dpsolomon, I must have missed your comment before I merged.  I'll open an issue to discuss this further.

@jrgnicho, I'll open an issue for your comment as well....although it's more straight forward.

Issue #12, opened 2014-10-18
Descartes package naming


shaun-edwards:
Currently, all source/include files are include in the `descartes_trajectory_planning` package.  I would like to propose a new package hierarchy to minimize dependencies and keep things better organized.
- descartes (repo)
  - descartes_core - All abstract and generalize-able classes, common utilities.  Basic implementations can be stored here until interfaces become defined (I'm thinking about the planner here).  This would basically be a rename of `descartes_trajectory_planning`.
  - descartes_moveit - This package would introduce a dependency on moveit.  Wrappers and class implementations that are moveit specific will be stored here.
  - descartes_gui - future - Common GUI tools for editing/managing paths, trajectory planning tools, etc..
  - descartes_planners - future - Specific implementation of planners

NOTE - we should also change namespaces to be identical to package names (this is the ROS convention).

shaun-edwards:
@jrgnicho & @dpsolomon, care to comment

jrgnicho:
Agreed, It's very likely that we'll have to rely on some parts of moveit at some point, thus separating the interfaces from the possible moveit dependent specializations is good forward thinking.  This will be specially useful if we end up having to implement a RobotModel subclass as a wrapper to moveit RobotState. I've already started working with the descartes types so we should address this renaming soon in order to make the transition to these new package names least painful.

shaun-edwards:
Fixed in #19

PR #15, opened 2014-10-26
Trajectory pt implementations

shaun-edwards:
This PR is dependent on #14.  It will capture the remaining items in PR #13.

### DIFF
--- a/descartes_trajectory_planning/include/descartes_trajectory_planning/trajectory_pt.h
+++ b/descartes_trajectory_planning/include/descartes_trajectory_planning/trajectory_pt.h
@@ -44,8 +44,6 @@ struct Frame
   Frame(){};
   Frame(const Eigen::Affine3d &a):
     frame(a), frame_inv(a.inverse()) {};
-  Frame(const Frame &a):
-    frame(a.frame), frame_inv(a.frame_inv) {};
 
   EIGEN_MAKE_ALIGNED_OPERATOR_NEW;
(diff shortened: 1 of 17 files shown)


shaun-edwards:
@dpsolomon, I'm manually porting your changes in the PR.  It's not done yet, but I wanted to make sure everyone knew I was working on it.

shaun-edwards:
This PR is ready for review.  It captures the remaining issues in PR #13.  It also lays the groundwork for a package rename #12 (namespace names corrected in a backwards compatible way).  Test methods were implemented to ensure that cartesian/joint points can be constructed.

One open issue (to be added) is the sampling of a waypoints and whether the waypoints or an external sampler should encapsulate that functionality.

@jrgnicho, please review.

descartes_trajectory_planning/include/descartes_trajectory_planning/trajectory_pt.h
```diff
@@ -44,6 +44,8 @@ struct Frame
   Frame(){};
   Frame(const Eigen::Affine3d &a):
     frame(a), frame_inv(a.inverse()) {};
+  Frame(const Frame &a):
```
jrgnicho: This copy constructor is equivalent to the default copy constructor provided by the compiler [here](http://www.cplusplus.com/articles/y8hv0pDG/).

jrgnicho:
Just that one minor observation.  Other than that looks good to me.
+1

PR #19, opened 2014-10-28
Descartes rename - Addresses issue #12

shaun-edwards:
Created `descartes_core` package.  @jrgnicho, please review

@ros-industrial-consortium/descartes - This will likely break any package that depends on Descartes.


jrgnicho:
+1

Issue #18, opened 2014-10-28
Sampling should be extracted from base trajectory types


shaun-edwards:
This issue is a result of discussions for PR #13 & #15.

At a high level a `trajectory_pt` should be independent of `solvers` and `samplers`.  Two key parts of the discussion are copied here for context:

@shaun-edwards

> The main rationale for making a trajectory_pt look like both a cartesian and joint point is to allow planners to operate on trajectories of mixed types. In order for this to work, a point needs to be able to calculate it's own IK/FK (and sample for that matter).

@dpsolomon 

> Each `TrajectoryPt` should contain a pointer to a sampler class, and implement a `bool sample(vector<JointPoint>&, int n)` method that simply calls the sampler with that point. The `sampler`should be able to incrementally sample according to its own discretization (or maybe discretization could also be passed as an arguement?). Repeated calls to `sample(.., 1) will return single-item vector, sample(.., 0) will return a vector of all joint solutions according to the discretization.
> The initial implementation should just call sample(.., 1) on each point, and the samplers should return the nominal joint solution. My though was that subsequent calls to sample() return results farther from the nominal until the tolerance is all used up.

dpsolomon:
+1 for getting back on solid footing

shaun-edwards:
@dpsolomon, a couple of questions
1. Is argument `n` the iteration number or the number of samples to return?  If it's not the iteration number, then I assume `sampler` will track it's own state.
2. What about sampling in cartesian/orientation space?  We will need to sample for cartesian points, although in this case, I am assuming that `getAllIK` will wrap a direct call to `ikfast` which will return _all_ joint solutions (big assumption on my part)
3. It feels like `samplers` will be specific to either cartesian or joint trajectory points.  Can you think of any other samplers?

dpsolomon:
> Is argument n the iteration number or the number of samples to return? If it's not the iteration number, then I assume sampler will track it's own state.

It is number of samples, and 0 means get all samples. ~~We might want to pass in discretization as well.~~ It will have to track its own state, but I don't expect it to keep track of previous samples. I would like the sampler to be (slightly) clever and be able to track multiple TrajectoryPts. Then we can wholesale change the discretization, or even the sample type for many/all points by reassigning the pointer (see below).  I have some notes for later about how this might happen.

[...]

PR #25, opened 2014-11-02
Initial implementation of JointTrajectoryPtDefaultSampler

dpsolomon:
DO NOT MERGE THIS PR!

This PR is for discussion only, it is not ready to be merged. Please let me know your thoghts on implementing a sampler. Some notes I already realize:
- I don't have a good way of passing data of different `TrajectoryPts` to the sampler
- I would like associate a sampler with each `TrajectoryPt` and implement `TrajectoryPt::sample()` which calls the pt's associated sampler, but I have to resolve some issues, including how two classes can easily depend on each other (sampler would depend on pt, pt would depend on sampler. Maybe its not a big deal if they remain in one `descartes_core` library?)
- The `SamplerState` currently holds alot of data as of the `init()` call. I would like to have the `operator++` actually calculate the next sample instead of just retrieving it from an array of precalculated samples.

I have included a unit test to show how the sampler may be used.

### DIFF
--- a/descartes_core/include/descartes_core/joint_trajectory_pt.h
+++ b/descartes_core/include/descartes_core/joint_trajectory_pt.h
@@ -86,6 +86,25 @@ class JointTrajectoryPt: public TrajectoryPt
 {
 public:
   EIGEN_MAKE_ALIGNED_OPERATOR_NEW;      //TODO is this needed when Frame already has it?
+
+public:
+  struct JointPointData
+  {
+    JointPointData():
+      tool(Eigen::Affine3d::Identity()), wobj(Eigen::Affine3d::Identity())
+    {};
+
+    JointPointData(const std::vector<TolerancedJointValue> &_joints,
+                   const Frame &_tool, const Frame &_wobj):
+                     joint_position(_joints),
+                     tool(_tool), wobj(_wobj)
+    {};
+
+    std::vector<TolerancedJointValue> joint_position;   /**<@brief Fixed joint position with tolerance */
+    Frame tool;                                         /**<@brief Transform from robot wrist to active tool pt. */
+    Frame wobj;                                         /**<@brief Transform from world to active workobject pt. */
+  };
+
 public:
   /**
     @brief Default joint trajectory point constructor.  All frames initialized to Identity, joint
@@ -146,6 +165,8 @@ public:
                                      const RobotModel &model,
                                      std::vector<double> &joint_pose) const;
 
+  virtual const void* getPointData() const;
+
   //TODO complete
   virtual void getJointPoses(const RobotModel &model,
                                std::vector<std::vector<double> > &joint_poses) const;
@@ -154,41 +175,34 @@ public:
   //TODO complete
   virtual bool isValid(const RobotModel &model) const;
 
-  //TODO complete
-  /**@brief Set discretization. Each joint can have a different discretization.
-   * @param discretization Vector of discretization values. If length=1, set all elements of discretization_ are set to value.
-   * @return True if vector is length 1 or length(joint_position_) and value[ii] are within 0-range(joint_position[ii]).
-   */
-  virtual bool setDiscretization(const std::vector<double> &discretization);
-
 inline
   void setJoints(const std::vector<TolerancedJointValue> &joints)
   {
-    joint_position_ = joints;
+    point_data_.joint_position = joints;
   }
 
   inline
   void setTool(const Frame &tool)
   {
-    tool_ = tool;
+    point_data_.tool = tool;
   }
 
   inline
   void setWobj(const Frame &wobj)
   {
-    wobj_ = wobj;
+    point_data_.wobj = wobj;
   }
   /**@} (end Setters section) */
-protected:
-  std::vector<TolerancedJointValue> joint_position_;  /**<@brief Fixed joint position with tolerance */
-  std::vector<double>               discretization_;  /**<@brief How finely to discretize each joint */
 
-  /** @name JointTrajectoryPt transforms. Used in get*CartPose() methods and for interpolation.
-   *  @{
-   */
-  Frame                         tool_;                  /**<@brief Transform from robot wrist to active tool pt. */
-  Frame                         wobj_;                  /**<@brief Transform from world to active workobject pt. */
-  /** @} (end section) */
+  virtual
+  bool setSampler(const TrajectoryPtSamplerPtr &sampler);
+
+  virtual
+  bool sample(size_t n);
+
+protected:
+
+  JointPointData point_data_;
 
 };
 
--- a/descartes_core/include/descartes_core/samplers/joint_trajectory_pt_default_sampler.h
+++ b/descartes_core/include/descartes_core/samplers/joint_trajectory_pt_default_sampler.h
@@ -1,0 +1,102 @@
+/*
+ * Software License Agreement (Apache License)
+ *
+ * Copyright (c) 2014, Dan Solomon
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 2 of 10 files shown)


dpsolomon:
This PR begins to address #18

descartes_core/include/descartes_core/joint_trajectory_pt.h
```diff
@@ -146,6 +146,8 @@ class JointTrajectoryPt: public TrajectoryPt
                                      const RobotModel &model,
                                      std::vector<double> &joint_pose) const;
 
+  virtual const void* getPointData() const;
```
shaun-edwards, line 168: ~~This method is the same in both `joint_trajectory_pt` and `cart_trajectory_pt`, so it could be in the base class.~~ ...just got that far.

descartes_core/include/descartes_core/samplers/joint_trajectory_pt_default_sampler.h
```diff
@@ -0,0 +1,101 @@
...
+protected:
+
+  struct SamplerState
+  {
+    typedef std::vector<double> JointSample;
+    std::vector<JointSample> samples;
+    std::vector<size_t> sample_idx;
+    bool done;
+
+    JointSample current_sample();
```
shaun-edwards, line 69: naming.

descartes_core/include/descartes_core/samplers/joint_trajectory_pt_default_sampler.h
```diff
@@ -0,0 +1,101 @@
...
+      samples.clear();
+      sample_idx.clear();
+      done = false;
+    }
+
+    /* prefix */
+    SamplerState& operator++ ();
+
+    /* postfix */
+    SamplerState operator++ (int)
```
shaun-edwards, line 82: I've never seen this before...makes sense.

descartes_core/include/descartes_core/samplers/joint_trajectory_pt_default_sampler.h
```diff
@@ -0,0 +1,101 @@
...
+    {
+      SamplerState tmp(*this);
+      ++(*this);
+      return tmp;
+    }
+  };
+
+  virtual
+  sensor_msgs::JointState sample1(SamplerState &sampler);
```
shaun-edwards: This brings in a dependency on `sensor_msgs`.  Is this what you want?  Is there a reason to not stick with `vector<double>`?

descartes_core/include/descartes_core/samplers/trajectory_pt_sampler.h
```diff
@@ -0,0 +1,69 @@
...
+   *
+   * Subsequent calls to this function return more solutions (if they remain. If sampler is complete, will return false).
+   *
+   * @param[out] solutions vector of JointStates.
+   * @param[in] n Number of samples to perform (not necessarily equal to number of solutions returned). n=0 means get all remaining solutions.
+   * @param[in] pt TrajectoryPt to sample from.
+   * @return True if new solutions were successfully created.
+   */
+  virtual
+  bool sample(std::vector<sensor_msgs::JointState> &solutions, size_t n, const TrajectoryPt &pt) = 0;
```
shaun-edwards: The ROS/Google convention is to put output parameters last (after input parameters).

descartes_core/include/descartes_core/trajectory_pt.h
```diff
@@ -176,6 +179,8 @@ class TrajectoryPt
 
 };
 
+typedef boost::shared_ptr<TrajectoryPt> TrajectoryPtPtr;
```
shaun-edwards: Might want to use this convenience macro, see [here](https://github.com/ros-industrial-consortium/descartes/blob/hydro-devel/descartes_core/include/descartes_core/utils.h#L27).

descartes_core/src/samplers/joint_trajectory_pt_default_sampler.cpp
```diff
@@ -0,0 +1,183 @@
...
+    {
+      ++sample_idx[joint_idx];
+      break;
+    }
+  }
+
+  return *this;
+}
+
+std::vector<double> JointTrajectoryPtDefaultSampler::SamplerState::current_sample()
```
shaun-edwards, line 72: method naming.

shaun-edwards:
> I don't have a good way of passing data of different TrajectoryPts to the sampler

I think this could become easier depending on the answer to the next few questions

> I would like associate a sampler with each TrajectoryPt and implement TrajectoryPt::sample() which calls the pt's associated sampler, but I have to resolve some issues, including how two classes can easily depend on each other (sampler would depend on pt, pt would depend on sampler. Maybe its not a big deal if they remain in one descartes_core library?)

Right now a single sampler manages all trajectory points, but it does make sense for a point to be able to sample itself.  Are we really saving that much by taking this approach, vs. embedding a sampler in each point.  I suppose changing sampler types would be faster with a single sampler instance, but it would really only save a `for` loop.

descartes_core/include/descartes_core/samplers/joint_trajectory_pt_default_sampler.h
dpsolomon: It depends on where we actually do IK for Cartesian points. It is conceivable that the sampler could know about the robot state and the `TrajectoryPt` remain purely process-descriptive.

dpsolomon:
> Right now a single sampler manages all trajectory points... 

A single sampler _could_ manage all `TrajectoryPts` of a particular type

> ..., but it does make sense for a point to be able to sample itself. Are we really saving that much by taking this approach, vs. embedding a sampler in each point. I suppose changing sampler types would be faster with a single sampler instance, but it would really only save a for loop.

Also changing sampler parameters. And the issue of embedding the sampler causing a complex interdependancy (see above).

I'm not really fixed on one idea or the other. The more important issue I see is that right now, if you have a `vector<TrajectoryPt>` with different point types in it, I don't see a good way in the code to iterate through and change anything about the sampler (or anything specific to the derived class, for that matter) because we will not know which type of `TrajectoryPt` it is. 

I almost feel like an individual sampler should be tied to each point, but a map external to both should track what point goes to what sampler (`map< TrajectoryPt::ID, boost::shared_ptr<TrajectoryPtSampler> >`). The obvious downside is it is a second object to track alongside the vector of points.

dpsolomon:
So I have a basic implementation of the `TrajectoryPt` calling its own sampler, but it feels very messy. I couldn't make it any nicer than two (2!) function prototypes that return `const void*`, and that just doesn't feel right. I'll push what I have soon, but if anyone has any ideas on how this might be implemented nicer, please let me know.

shaun-edwards:
It helps to think about how the sampler will be used instead of the data structures and specifics behind the implementation.  First, let's just consider `TrajectoryPt`.  We want to be able to get new samples from the trajectory point.  We currently have `get*Poses` methods for full sampling.  In addition, it looks like we want to add a `getNext*Pose` method to support incremental sampling.  A `resetSampling` method could also be useful.  At a `TrajectoryPt` level, is there any other functionality that is required?  Do we need to have _direct_ access to the sampler at the `TrajectoryPt` level? (I think the answer is no).

Once we can agree on this, we can start to look at how the specific implementations of `TrajectoryPt` (i.e. joint & cartesian) implement these methods.

dpsolomon:
I'm OK with the interface we have.  My question was directed at the
implementation. I pushed what I have currently this morning; You can look
at the test to see how annoying it could be to keep the interface generic.
One thought I have is to keep the generic interface, and add a simple
interface specific to the sampler type. For instance, a `JointPtSampler`
could have a generic interface inherited form the sampler base class, but
it could also have a specific joint sample interface; Any derived classes
of the `JointPtSampler` could also use the specific (and hopefully simpler)
interface.

shaun-edwards:
My point with the comment above is that a generic (`TrajectoryPt` level) sampler is not required.  I agree with having a generic Sampler base class with Joint/Cartesian versions that refine it (a base class really isn't required, but a hierarchy might be useful).  Using `<static_cast>` should be avoided if at all possible.  

Here's my suggested approach (you don't have to do this, but it's what I am thinking):
1. Remove any mention of samplers from `TrajectoryPt`.  Add in methods that return the results of a sampler (just not any implementation)
2. Create Joint/Cartesian sampler interface definitions.  These do not have to inherit from a base type (unless of course some obvious commonality is present ).  Any method returning a `void*` that must be recast isn't common.
3. Create default(uniform) Joint/Cartesian samplers that are passed to specific trajectory points on construction.

Maybe a short/long phone call would be easier ;)

dpsolomon:
Replaced by #40

PR #28, opened 2014-11-05
Fleshing out BGL solver (add, modify, delete, and solve)

psturgeon:
This PR is incomplete at this point but I want to make it available for review and comments.
- The main change was converting getShortestPath to automatically determine the beginning and ending graph vertices and completing the  conversion to using boost:UUIDs.
- modifyTrajectory and removeTrajectory are not implemented yet.
- addTrajectory is very rough and incomplete.
- The graph was converted to a bidirectional graph in order to simplify identification of the start vertex/vertices; edges are still only created as if they were in a directed graph and should not effect the solution (have not done any load testing yet to compare performance of the bidirectional graph vs the direction graph).


descartes_core/include/descartes_core/planning_graph.h
```diff
@@ -83,12 +85,19 @@ class PlanningGraph
...
   bool insertGraph(std::vector<TrajectoryPtPtr> *points);
 
-  // TODO: addTrajectory
-  // TODO: modifyTrajectory
-  // TODO: removeTrajectory
+  /** @brief adds a single trajectory point to the graph
+   * @param point The new point to add to the graph
+   * @return True if the point was successfully added
+   */
+  bool addTrajectory(TrajectoryPtPtr point, TrajectoryPt::ID previous_id, TrajectoryPt::ID next_id);
```
shaun-edwards, line 97: Do you mean to pass a copy or can you pass by reference?

descartes_core/src/planning_graph.cpp
```diff
@@ -111,48 +116,214 @@ bool PlanningGraph::insertGraph(std::vector<TrajectoryPtPtr> *points)
...
-      weight_map(get(&JointEdge::transition_cost, dg_)).distance_map(
-          boost::make_iterator_property_map(weights.begin(), get(boost::vertex_index, dg_))).predecessor_map(
-          &predecessors[0]));
+  // TODO: get joint poses from new trajectory point (unique id?) and
+  //      add to trajectory_point_to_joint_solutions_map_ and to joint_solutions_map_
+
+  // TODO: calculate weights from id_previous to id
+  // TODO: calculate weights from id to id_next
+  /*
+   for (std::list<int>::iterator start_joint_iter = start_joint_ids.begin(); start_joint_iter != start_joint_ids.end();
```
shaun-edwards: Indentation is messed up for a while. <EDIT>....Oh, it's in a comment.  Please use `//` for comments...it's easier to recognize in the github diff viewer.

descartes_core/src/planning_graph.cpp
```diff
@@ -111,48 +116,214 @@ bool PlanningGraph::insertGraph(std::vector<TrajectoryPtPtr> *points)
...
 
-    // Starting from the destination point step through the predecessor map
-    // until the source point is reached.
-    while (current != start_id)
+    std::pair<InEdgeIterator, InEdgeIterator> in_ei = boost::in_edges(jv, dg_);
+    if (in_ei.first == in_ei.second)
     {
-      current = predecessors[current];
-      path.push_front(current);
+      std::cout << "START NODE: " << jv << "\n";
```
shaun-edwards: Should be a log message.

descartes_core/include/descartes_core/planning_graph.h
```diff
@@ -131,6 +140,10 @@ class PlanningGraph
   // map from Cartesian Point ID to applicable joint solutions per point
   std::map<TrajectoryPt::ID, std::list<TrajectoryPt::ID> > trajectory_point_to_joint_solutions_map_;
 
+  bool findStartVertices(std::list<int> *start_points);
```
shaun-edwards: This isn't important for this PR, but we generally don't use pointers unless they are required.  If the argument can be passed by reference, then that is desired.  At the very least, it means you don't have to check for `NULL` in your function.  These functions are protected, so it's not critical for this PR.

shaun-edwards:
+1...let's get this integrated with @jrgnicho

PR #29, opened 2014-11-10
Refactored RobotModel test code to be parameterized gtest

shaun-edwards:
Pulled `RobotModel` interface test into `descartes_core`.  Tests can not be pulled into any implementation of `RobotModel` (`descartes_moveit` and others).

### DIFF
--- a/descartes_moveit/test/moveit_state_adapter_test.cpp
+++ b/descartes_moveit/test/moveit_state_adapter_test.cpp
@@ -16,78 +16,52 @@
  * limitations under the License.
  */
 
-#include "descartes_core/pretty_print.hpp"
 #include "descartes_moveit/moveit_state_adapter.h"
+#include "descartes_core_test/robot_model_test.hpp"
 #include "moveit/robot_model_loader/robot_model_loader.h"
-#include "console_bridge/console.h"
-#include "ros/console.h"
 #include <gtest/gtest.h>
-#include <iostream>
 
 using namespace descartes_moveit;
+using namespace descartes_core;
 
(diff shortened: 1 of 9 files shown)


descartes_moveit/test/moveit_state_adapter_test.cpp
```diff
@@ -16,78 +16,52 @@
...
+// This variable must be global in order for the test to pass.
+// Destructing the robot model results in a boost mutex exception:
+// ---
+// /usr/include/boost/thread/pthread/pthread_mutex_scoped_lock.hpp:26:
+// boost::pthread::pthread_mutex_scoped_lock::pthread_mutex_scoped_lock(pthread_mutex_t*):
+// Assertion `!pthread_mutex_lock(m)' failed.
+// ---
 
-  // virtual void TearDown() {}
+robot_model_loader::RobotModelLoaderPtr robot_;
```
shaun-edwards, line 39: @dpsolomon, see the comment above this declaration.  Does this sound familiar?  This issues shouldn't hold up this PR (it's in test code), but I could not figure it out for the life of me.

shaun-edwards:
@jrgnicho, please review.

jrgnicho:
+1

descartes_moveit/test/moveit_state_adapter_test.cpp
dpsolomon, line 39: I thought `TearDown()` just overrode the `::testingTest` class function, so I have no idea why it would do this. I didn't see that you were doing any locks in your stuff, so I think it has to be a throw from the google test framework.

PR #33, opened 2014-11-12
Moveit adaptor fixes

jrgnicho:
- transforms goal tool pose to robot base before calling ik
  - JointTrajectoryPt returns a single joint solution in the 'getPoses(...)' method

@shaun-edwards  please review

### DIFF
--- a/descartes_core/src/joint_trajectory_pt.cpp
+++ b/descartes_core/src/joint_trajectory_pt.cpp
@@ -99,7 +99,9 @@ bool JointTrajectoryPt::getNominalJointPose(const std::vector<double> &seed_stat
 void JointTrajectoryPt::getJointPoses(const RobotModel &model,
                                       std::vector<std::vector<double> > &joint_poses) const
 {
-  joint_poses.clear();
+  std::vector<double> empty_seed;
+  joint_poses.resize(1);
+  getNominalJointPose(empty_seed,model,joint_poses[0]);
 }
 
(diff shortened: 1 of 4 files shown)


shaun-edwards:
@jrgnicho, I'm going to need you to walk me through this one.

jrgnicho:
To clarify this PR's changes,
The  moveit::core::RobotState::setFromIK(...) method assumes that the pose is defined relative to the robot model root frame/link.  In most cases, the root frame and the desired world frame are the same but for cases where these frames differ then a solution will not be found or will be incorrect at best.  The changes made here apply a transform to the  tool pose so that it is defined relative to the user defined world frame.  Thus, the new 'world_to_root_' member in the MovetStateAdaptor captures the transform from the user designated 'world frame' to the 'robot model root frame' and vice versa.

jrgnicho:
The last two commits [314da5a](https://github.com/ros-industrial-consortium/descartes/commit/314da5ad0bcd6670bf0c5d929f9707cea6449762) and [85aac17](https://github.com/ros-industrial-consortium/descartes/commit/85aac17368ff2a8d5605b2d873a1785eec0a56ca) removed the local ik related members.  A quick test was made to verify that the default timeout and attempt values in the kinematics.yaml file were used.  This test was a simple log message which was removed afterwards and was not included in these commits.
@shaun-edwards  please review

shaun-edwards:
+1... @psturgeon, please be sure to update your local repo with the additional changes.

Issue #38, opened 2014-12-03
Failure to find valid solutions for sampled cartesian points


jrgnicho:
Valid solutions for the sampled cartesian points can not be produced.

jrgnicho:
@shaun-edwards  can this be closed?

shaun-edwards:
@jrgnicho - please try to include the short description of the PR.  Then the issues will automatically close.

PR #39, opened 2014-12-03
fix to issue #38 failure to solved sampled cartesian points

jrgnicho:
This PR addresses issue #38 
@shaun-edwards please review

### DIFF
--- a/descartes_core/src/cart_trajectory_pt.cpp
+++ b/descartes_core/src/cart_trajectory_pt.cpp
@@ -52,9 +52,9 @@ namespace descartes_core
 
   if(orient_increment > 0)
   {
-    nrx = (frame.orientation_tolerance.x_upper - frame.orientation_tolerance.x_lower)/orient_increment + 1;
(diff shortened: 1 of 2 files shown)


descartes_core/src/cart_trajectory_pt.cpp
```diff
@@ -102,7 +102,7 @@ namespace descartes_core
             for(size_t nn = 0; nn < ntz; ++nn)
             {
               tz = frame.position_tolerance.z_lower + pos_increment * nn;
-              sampled_frame = Eigen::Translation3d(tx,ty,tz) *
+              sampled_frame = frame.frame*Eigen::Translation3d(tx,ty,tz) *
```
shaun-edwards: This isn't what I intended.  `frame.frame` is the nominal value, but any frame within the tolerances is valid (the tolerance values `upper/lower` should encompass the nominal).  Debug log messages ([here](https://github.com/ros-industrial-consortium/descartes/blob/hydro-devel/descartes_core/include/descartes_core/cart_trajectory_pt.h#L106-L109) should show this).

shaun-edwards:
+1

PR #40, opened 2014-12-09
Initial implementation (take 2) of TrajectoryPtSamplers

dpsolomon:
DO NOT MERGE THIS PR!

This PR is for discussion only, it is not ready to be merged. Please let me know your thoughts on implementing a sampler.   This PR beings to address #18.

Some differences from my last attempt (https://github.com/dpsolomon/descartes/tree/samplers):
- Each TrajectoryPt type has its own sampler, e.g. a JointSampler for a JointTrajectoryPt.
- Each sampler type has a pure base class.

### DIFF
--- a/descartes_core/include/descartes_core/cart_trajectory_pt.h
+++ b/descartes_core/include/descartes_core/cart_trajectory_pt.h
@@ -25,9 +25,11 @@
 #ifndef CART_TRAJECTORY_PT_H_
 #define CART_TRAJECTORY_PT_H_
 
+#include <eigen_stl_containers/eigen_stl_vector_container.h>
 #include <moveit/kinematic_constraints/kinematic_constraint.h>
+#include <ros/console.h>
 #include "descartes_core/trajectory_pt.h"
-#include "ros/console.h"
+#include "descartes_core/samplers/cart_pt_sampler_base.h"
 
 typedef boost::shared_ptr<kinematic_constraints::PositionConstraint> PositionConstraintPtr;
 typedef boost::shared_ptr<kinematic_constraints::OrientationConstraint> OrientationConstraintPtr;
@@ -168,6 +170,28 @@ struct TolerancedFrame: public Frame
   OrientationTolerance          orientation_tolerance;
   PositionConstraintPtr         position_constraint;
   OrientationConstraintPtr      orientation_constraint;
+
+  Eigen::Vector3d upperPositionBound() const
+  {
+    return frame.translation() + Eigen::Vector3d(position_tolerance.x_upper, position_tolerance.y_upper, position_tolerance.z_upper);
+  }
+
+  Eigen::Vector3d lowerPositionBound() const
+  {
+    return frame.translation() - Eigen::Vector3d(position_tolerance.x_lower, position_tolerance.y_lower, position_tolerance.z_lower);
+  }
+
+  Eigen::Vector3d upperRPYBound() const
+  {
+    Eigen::Vector3d rpy = frame.rotation().eulerAngles(2,1,0);
+    return Eigen::Vector3d(rpy(2), rpy(1), rpy(0)) + Eigen::Vector3d(orientation_tolerance.x_upper, orientation_tolerance.y_upper, orientation_tolerance.z_upper);  //TODO check that this is correct
+  }
+
+  Eigen::Vector3d lowerRPYBound() const
+  {
+    Eigen::Vector3d rpy = frame.rotation().eulerAngles(2,1,0);
+    return Eigen::Vector3d(rpy(2), rpy(1), rpy(0)) - Eigen::Vector3d(orientation_tolerance.x_lower, orientation_tolerance.y_lower, orientation_tolerance.z_lower);  //TODO check that this is correct
+  }
 };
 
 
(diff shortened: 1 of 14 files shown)


dpsolomon:
There is currently a short JointSampler test and 70% of a CartSampler test. I will finish the Cartesian sampler integration, but most of this can still be reviewed as is.  @shaun-edwards Do we only care about yaw sampling for the moment?

shaun-edwards:
FYI...going to be later in the week when I get to this.

descartes_core/include/descartes_core/samplers/cart_pt_sampler_base.h
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * Sample increment describes how to increment the value of each sample.
+ * For a Cartesian sampler, 6 sampling increment values are needed, 3 for position and 3 for rotation.
+ * It is left up to the implementation as to how the rotation increments are used.
+ * @{
+ */
+virtual
+std::vector<double> getSampleIncrement() const = 0;
+virtual
+void setSampleIncrement(const std::vector<double> &values) = 0;
```
shaun-edwards, line 60: Was this purposely left ambiguous (as a vector)?  Could a more explicit position/orientation set of parameters be used?

descartes_core/include/descartes_core/cart_trajectory_pt.h
```diff
@@ -298,6 +352,9 @@ class CartTrajectoryPt : public TrajectoryPt
   double                        pos_increment_;    /**<@brief Sampling discretization in cartesian directions. */
   double                        orient_increment_; /**<@brief Sampling discretization in angular orientation. */
 
+  CartPtSamplerBasePtr          tool_sampler_,  /**<@brief Cartesian sampler for tool frame. */
```
shaun-edwards, line 355: Is this safe/desired?  By using a pointer, the same sampler could be assigned to multiple points.  I'm not sure this is ever desired, given the sampler holds it's own sample state, nominal, and bound values.

descartes_core/include/descartes_core/joint_trajectory_pt.h
```diff
@@ -190,6 +215,8 @@ inline
   Frame                         wobj_;                  /**<@brief Transform from world to active workobject pt. */
   /** @} (end section) */
 
+  JointPtSamplerBasePtr         sampler_;               /**<@brief Joint Pt Sampler */
```
shaun-edwards, line 218: See comment about CartesianPtSampler above.

descartes_core/include/descartes_core/samplers/cart_trajectory_pt_default_yaw_sampler.h
```diff
@@ -0,0 +1,64 @@
...
+  bool initPositionData(const Eigen::Affine3d &nominal,
+                        const Eigen::Vector3d &upper_position_bound = Eigen::Vector3d::Zero(), const Eigen::Vector3d &lower_position_bound = Eigen::Vector3d::Zero(),
+                        const Eigen::Vector3d &upper_rpy_bound = Eigen::Vector3d::Zero(), const Eigen::Vector3d &lower_rpy_bound = Eigen::Vector3d::Zero());
+
+  virtual
+  bool sample(Eigen::Affine3d &result);
+
+protected:
+
+  /* Private Members */
```
shaun-edwards, line 54: Actually they are protected ;)

descartes_core/include/descartes_core/samplers/joint_trajectory_pt_default_sampler.h
```diff
@@ -0,0 +1,87 @@
...
+    std::vector<size_t> sample_idx;             /**<@brief An index (into @e samples) of the current sample to be taken for each joint. */
+    bool done;                                  /**<@brief Flag set when all samples have been taken. */
+
+    /**@brief Retrieve current sample by getting each sample value by corresponding sample index.
+     * @return Joint position (empty vector if @e done is set.
+     */
+    std::vector<double> current_sample();
+
+    /**@brief Reset the SampleState. */
+    void clear();
```
shaun-edwards, line 68: Should this be called `reset()`?  Regardless, should a `reset()` method be added to the base samplers?.

descartes_core/src/cart_trajectory_pt.cpp
```diff
@@ -253,10 +253,73 @@ bool CartTrajectoryPt::isValid(const RobotModel &model) const
   return model.isValid(robot_pose);
 }
 
+EigenSTL::vector_Affine3d CartTrajectoryPt::sample(size_t n, CartPtSamplerBasePtr &sampler)
```
shaun-edwards, line 256: What is the difference between this sample call and the tool/wobj sample calls below?
Is calling a sample method with a sampler argument safe?  What if the sampler is initialized with a nominal/bounds that isn't at all related to the specific point.

shaun-edwards:
@dpsolomon, I'm only about half way through reviewing, but I think it might help to address my comments so far.  In general, the samplers capture similar position/bounds information as the points themselves.  This could cause problems if the points and samplers are not kept in sync.  
I was expecting to see modifications to the `TrajectoryPt` base class, but I didn't not see it.  How are the samplers supposed to be called from a `TrajectoryPt` level (maybe they are expected to be called from the existing methods)?

What if we use some templating tricks to achieve what I think your trying to achieve.  Similar to how the STL containers allow you to override allocators (if you are badass enough to do such a thing).

shaun-edwards:
@dpsolomon, are you still working on this or should we close it?

dpsolomon:
I haven't touched it in quite some time. It looks like the project has moved far past this, I would just close it out.

Issue #41, opened 2014-12-11
PlanningGraph destructor cause crashs


jrgnicho:
The destructor of the PlanningGraph may attempt to delete an unassigned non null pointer if no points were ever added to the planner.

jrgnicho:
Addressed by PR #42

PR #42, opened 2014-12-11
initializes the cartesian map pointer to null

jrgnicho:
This is a fix to issue #41 
@shaun-edwards please review


shaun-edwards:
+1

Issue #43, opened 2014-12-18
Closest Joint Pose Implementation


jrgnicho:
The CartTrajectoryPt::getClosestJointPose(...) methods needs to be implemented in order to address the sparse planner implementation discussed in issue #37.

jrgnicho:
Current implementation in #44 only works for simple cases.  A more generalized implementation (perhaps quaternion based approach) is needed

shaun-edwards:
This issue still continues to be a problem.  Current implementations do not generalize to a fast method for the `CartesianPt` base class.  This is partly due to translation and rotation representation and specifically continuity of representation.  A potential solution to this may be the Dual-Quaternions representation.  See references below:
- http://wscg.zcu.cz/wscg2012/short/a29-full.pdf
- https://en.wikipedia.org/wiki/Screw_theory
- https://en.wikipedia.org/wiki/Dual_quaternion#More_on_spatial_displacements

Issue #37, opened 2014-11-30
Graph planning algorithm is slow when solving for large number (1000's) of points


shaun-edwards:
The current (working) graph search algorithm is slow when a trajectory with 1000's of points is desired.  Such trajectories could be common, especially when trying to control positions with high accuracy or when performing linear motion with a manipulator.  Preliminary tests indicate that graph construction is the expensive(time) operation.

For example, assume the following: 1000 Cartesian trajectory points specified, with a free yaw rotation (at 1 degree resolution - required for smooth motion), on a standard 6 DOF industrial robot, w/o collision checking.  The number of IK calls (approximately the same as number of nodes) in the graph (i.e. that must be constructed) is:

1000 Cart pts \* 360 yaw samples/Cart pt \* 10 IK solutions (assuming KDL) = 3.6M (IK solver calls)

Current tests have been executed without yaw sampling and take several seconds to load.  Even a sparse yaw sampling (which isn't good for motion smoothness) will result in very long graph construction times.

An obvious improvement to the existing algorithm is to use IKFast instead of KDL.  In the past we have seen a 10x faster solve time with IKFast.  Our results may be better given IKFast doesn't need to "search" for multiple IK solutions.  It can calculated them all directly.  

While this is an obvious change, it will not solve the main problem: The current planning algorithm does not scale with time.  Furthermore, simple planning problems take as long as to solve as more complex ones.

shaun-edwards:
I propose an adaptive trajectory sampling that starts with a sparse set of trajectory points and interpolates between sparse points, while still maintaining the dense trajectory constraints.  If any dense trajectory constraint is not achievable given the sparse plan, the unachievable point is added to the trajectory and resolved.

Here is the algorithm in pseudo-code

```
For each N sub-samples in trajectory:
  Add sub-sample to graph //uses existing graph planner API
  Solve graph
For each P trajectory-pt in trajectory
  If ( trajectory-pt.getClosestJointPose(interpolated joint pose, model) ) //must be implemented in trajectory pt
     Add joint pose to solution
  Else
     Add point to graph //lightweight operation in current solver
     (Re) Solve graph
     Reset loop

```

Under a best case scenario (i.e. the sparse trajectory results in a solution for the dense trajectory), the number of IK solutions can be calculated as follows, with the following assumptions:
- the required yaw sampling is also less dense (every 10 degrees or further) since trajectory points will be interpolated between sparse points
[...]

PR #44, opened 2014-12-19
Issue #43 closest joint pose implementation

jrgnicho:
This PR addresses issue #43 as follows:
- Implemented CartesianPt::getClosestJointPose(...) with the following approach:
  1. Computes FK using the seed_pose passed to the method.
  2. Computes the x,y ,z, rx, ry,rz from the pose obtained from FK.  The orientation values are computed with the Eigen::eulerAngles(...) method
  3. Each pose value is then checked to verify that it is within the points tolerance.  If it isn't then that value is set to one of the bounds for the points pose and ik is solved for the modified pose values
- Created Unit test to validate the method described above.

Note:  Since the Eigen::getEuler method may return different rpy combination for slightly different poses this approach may occasionally yield orientation values that  are outside the cartesian point tolerances.  Thus, the closest joint pose returned in these cases could be really far from the seed pose when in practice there could be a much closer one.

@shaun-edwards , @dpsolomon, any suggestions will be very welcomed.

### DIFF
--- a/descartes_core/CMakeLists.txt
+++ b/descartes_core/CMakeLists.txt
@@ -2,6 +2,7 @@ cmake_minimum_required(VERSION 2.8.3)
 project(descartes_core)
 
 set(CMAKE_BUILD_TYPE Debug)
+add_definitions(-std=gnu++0x)
 
 find_package(catkin REQUIRED COMPONENTS
   console_bridge
--- a/descartes_core/include/descartes_core/cart_trajectory_pt.h
+++ b/descartes_core/include/descartes_core/cart_trajectory_pt.h
@@ -70,8 +70,8 @@ struct ToleranceBase
     #param tol Total tolerance zone (assumed symetric about nominal)
     */
   template<typename T>
-  static T createSymmetric(const double x, const double y, const double z,
-                                          const double tol)
+  static T createSymmetric(const double& x, const double& y, const double& z,
+                                          const double& tol)
   {
     return(createSymmetric<T>(x, y, z, tol, tol, tol));
   }
@@ -81,7 +81,7 @@ struct ToleranceBase
     @param x, y, z
     */
   template<typename T>
-  static T zeroTolerance(const double x, const double y, const double z)
+  static T zeroTolerance(const double& x, const double& y, const double& z)
   {
     return(createSymmetric<T>(x, y, z, 0.0, 0.0, 0.0));
   }
@@ -174,8 +174,8 @@ struct TolerancedFrame: public Frame
     orientation_tolerance = ToleranceBase::createSymmetric<OrientationTolerance>(rpy(2),rpy(1),rpy(0),0);
   };
 
-  TolerancedFrame(const Eigen::Affine3d &a, const PositionTolerance pos_tol,
-                  const OrientationTolerance orient_tol) :
+  TolerancedFrame(const Eigen::Affine3d &a, const PositionTolerance &pos_tol,
+                  const OrientationTolerance &orient_tol) :
     Frame(a), position_tolerance(pos_tol), orientation_tolerance(orient_tol){}
 
   PositionTolerance             position_tolerance;
--- a/descartes_core/include/descartes_core/pretty_print.hpp
+++ b/descartes_core/include/descartes_core/pretty_print.hpp
@@ -197,8 +197,8 @@ namespace pretty_print
             if (delimiters_type::values.prefix != NULL)
                 stream << delimiters_type::values.prefix;
 
-            if (begin(_container) != end(_container))
-            for (TIter it = begin(_container), it_end = end(_container); ; )
+            if (pretty_print::begin(_container) != pretty_print::end(_container))
+            for (TIter it = pretty_print::begin(_container), it_end = pretty_print::end(_container); ; )
             {
                 stream << *it;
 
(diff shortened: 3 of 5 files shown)


ClayFlannigan:
You should use quaternions when comparing angles rather than Euler angles.  Quaternions won't change in discontinuous ways like Euler angles.  You can take the dot product of two quaternions to compare the cosine of the angle between them.  Thus two quaternions are approximately equivalent if q1.dot(q2) > 1 - tolerance.  Or more directly q1.angularDistance(q2) < tolerance.

shaun-edwards:
@jrgnicho, is this an issue if only one euler angle is ever sampled (i.e. our current use case)?

jrgnicho:
@ClayFlannigan thank you for your input.  I think the quaternion approach might address that issue if we are setting non-zero tolerances on one euler angle only.  Since we only have tolerances on the wrist rotation then the quaternion approach should be sufficient.

@shaun-edwards this problem could happen regardless of how we sample the trajectory point.

shaun-edwards:
This problem is a lot more complex than I originally thought.  For simple constraints (i.e. axially-symmetric cartesian points) the problem is simplified greatly.  For the near term we will address this in a class that inherits from the more general `CartesianTrajPt` for this simpler application.

@jrgnicho, I will cherry pick some of your commits in this PR as needed (I may have some questions on a few changes).

As for the more general solution, @ClayFlannigan is right in that we need to work in quaternion space.  This [article](http://www.csse.uwa.edu.au/~du/ps/Huynh-JMIV09.pdf) provides some interesting perspective on the different representations between rotations, as well as relative complexity.  The paper focuses on calculating the difference between rotations, not solving for the nearest rotation as required by our function.  

To solve the problem of finding the nearest rotation within a certain range for a given pose, it helps to look at the simpler 3D translation problem.  For a 3D (x,y,z) coordinate, the closes pose within a certain range is found by simply limiting the x,y,z components to the pose range.  This assumes that the pose range defines a rectangular (volumetric) tolerance zone, which it does.  Finding the nearest rotation could be as simple if it defined a similar rectangular volumetric tolerance zone.  Unfortunately it does not.  Rather it defines an unknown(at least to me) tolerance zone on the surface of a hypersphere in 4D space (something that I cannot visualize).  While this is harder to visualize, the math (in a generalized manor) should be exactly to the same as in the 3D translation problem.  We simply need to find the boundary defined by the orientation tolerance range that is closest to the passed in pose.  If we solve the 3D translation problem in a generalized manor (i.e. pure vector math, instead of simply truncating coordinate values), then we will have the answer in 4D.

shaun-edwards:
This [eigen bug report](http://eigen.tuxfamily.org/bz/show_bug.cgi?id=609) seems to indicate Eigen returns euler angles consistently (in cases of gimble lock, the first angle is set to zero, so it's still consistent even if not unique).  It may be possible to utilize Eigen to transform any Euler angles into a consistent set (even max/min ranges) and then internally use only consistent euler angles.  Could it be possible that angle ranges are not continuous in consistent Eigen euler angles (i.e. a range maps to a discontinuous set 0-01 and 3-PI)?  Even if this occurs, we could catch this case, limit the range, and provide a warning.  This would arbitrarily limit a range, albeit with a potentially simpler implementation.

dpsolomon:
WRT finding the closest angles in quaternion space, I agree it's going to
be more difficult than the 3D position case. The general solution to
finding the closest solution is something like projecting the point along
the normal of the bounding surface.

With independent tolerances (x, y, z tolerances that are functions of AT
MOST themselves) in Cartesian space, the bounding volume describes a
rectangular prism. Finding the closest solution to a point outside that
boundary is simple, because each variable can be independently shifted to
its closest extreme value.

Even in Cartesian space, finding the closest solution of a generalized
tolerance volume can be difficult. A sphere is an easy degenerate case, but
a cylindrical shape starts to expose some of the difficulty there. Any
polyhedron would present some difficulty in determining how to project the
point.

I think that one of the key problems with rotation space is that the
variables are not independent, and therefore a tolerance zone based on
rotations does not have an easily defined bounding volume (or at least not
one that I personally can imagine describing). I agree that if we can solve
the general case of a tolerance zone with dependant variables in 3D, the
math may extend to rotation space, or even quaternion space, but I do know
that it is not straightforward.

dpsolomon:
After looking at this a bit further, I realized I misstated that rotation space in general (SO3) does not have independant variables, when in fact it does (e.g. fixed rotations taken about world axes).  Therefore, it may be valid to simply restrict rotation angles to their nearest extreme limit **IF** our tolerances are defined by fixed rotations (I'm not 100% sure about this).

I agree with @ClayFlannigan  that we really should be working in quaternion space rather than Euler or fixed rotation denotations.

As an aside, what is the use-case for `closestPoint()`?  While this is a very interesting topic, is it the right approach this problem? I believe the sampling strategy we came up with doesn't require this method...

shaun-edwards:
This method is used for the sparse trajectory planner (see [here](https://github.com/swri-robotics/descartes/blob/amos_integration/descartes_core/src/sparse_planner.cpp#L555)) (note: we forked the repo in order to meet a project deadline.  We intend on doing a formal PR after delivery).  The plan for the sparse planner was outlined in PR #37.  The `closest` methods might also be helpful for an A\* search (if we every get one) as well.  The sparse planner appears to be working and is certainly faster.

A _key_ requirement of the `closest` methods is speed.  In particular, we are trying to return a joint solution without having to do to many IK calls (or sampling for that matter).

@jrgnicho PR seems to work well in that it returns a valid solution quickly, although sometimes it may fail to return a valid solution due to the issues with euler angles.  This is OK for the sparse search method, because it defaults to a search/sample based approach when the `closest` method fails.

PR #45, opened 2014-12-31
Sparse trajectory planner

jrgnicho:
This is PR addresses the issue #37 
- Added the SparsePlanner class for planning on a subset of points taken from a "denser" trajectory.
- It first generates a graph plan for the sparse set and then it interpolates through the remaining points in the 
  dense trajectory that were not included in the sparse set.
- It inherits from PlanningGraph and provides a similart interface as its base class
- Its interface includes methods for adding, modifying and removing points.
- Unit test programs loads a fake trajectory with 1000 points.

In order to run unit test run the following in a terminal:

```
catkin_make run_tests_descartes_core_gtest_descartes_core_sparse_planner_utest
```

@shaun-edwards  please review

### DIFF
--- a/descartes_core/CMakeLists.txt
+++ b/descartes_core/CMakeLists.txt
@@ -63,7 +63,8 @@ if(CATKIN_ENABLE_TESTING)
       test/cart_trajectory_pt.cpp
       test/joint_trajectory_pt.cpp
       test/cartesian_robot.cpp
-      test/cartesian_robot_test.cpp)
+      test/cartesian_robot_test.cpp
+      test/sparse_planner.cpp)
   catkin_add_gtest(${PROJECT_NAME}_utest ${UTEST_SRC_FILES})
   target_link_libraries(${PROJECT_NAME}_utest descartes_core)
 endif()
--- a/descartes_core/CMakeLists.txt
+++ b/descartes_core/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 2.8.3)
 project(descartes_core)
 
 set(CMAKE_BUILD_TYPE Debug)
-
+add_definitions(-std=gnu++0x)
 find_package(catkin REQUIRED COMPONENTS
   console_bridge
   moveit_core
@@ -44,11 +44,10 @@ add_library(descartes_core
             src/joint_trajectory_pt.cpp
             src/trajectory.cpp
             src/planning_graph.cpp
+            src/sparse_planner.cpp
             src/robot_model.cpp
 )
-##Disabling planning graph due to issues with
-##Trajectory point ID
-##  src/planning_graph.cpp
+
 target_link_libraries(descartes_core
                       ${catkin_LIBRARIES}
 )
--- a/descartes_core/CMakeLists.txt
+++ b/descartes_core/CMakeLists.txt
@@ -64,7 +64,6 @@ if(CATKIN_ENABLE_TESTING)
       test/joint_trajectory_pt.cpp
       test/cartesian_robot.cpp
       test/cartesian_robot_test.cpp)
-      #test/sparse_planner.cpp)
   catkin_add_gtest(${PROJECT_NAME}_utest ${UTEST_SRC_FILES})
   target_link_libraries(${PROJECT_NAME}_utest descartes_core)
 
--- a/descartes_core/CMakeLists.txt
+++ b/descartes_core/CMakeLists.txt
@@ -63,8 +63,15 @@ if(CATKIN_ENABLE_TESTING)
       test/cart_trajectory_pt.cpp
       test/joint_trajectory_pt.cpp
       test/cartesian_robot.cpp
-      test/cartesian_robot_test.cpp
-      test/sparse_planner.cpp)
+      test/cartesian_robot_test.cpp)
+      #test/sparse_planner.cpp)
   catkin_add_gtest(${PROJECT_NAME}_utest ${UTEST_SRC_FILES})
   target_link_libraries(${PROJECT_NAME}_utest descartes_core)
+
+  set(UTEST_SPARSE_PLANER_SRC_FILES test/utest.cpp
+      test/cartesian_robot.cpp
+      test/sparse_planner.cpp)
+  catkin_add_gtest(${PROJECT_NAME}_sparse_planner_utest ${UTEST_SPARSE_PLANER_SRC_FILES})
+  target_link_libraries(${PROJECT_NAME}_sparse_planner_utest descartes_core)
+
 endif()
--- a/descartes_core/include/descartes_core/robot_model.h
+++ b/descartes_core/include/descartes_core/robot_model.h
@@ -89,6 +89,12 @@ public:
   virtual bool getFK(const std::vector<double> &joint_pose, Eigen::Affine3d &pose) const = 0;
 
   /**
+   * @brief Returns number of DOFs
+   * @return Int
+   */
+  virtual int getDOF() const = 0;
+
+  /**
    * @brief Performs all necessary checks to determine joint pose is valid
    * @param joint_pose Pose to check
    * @return True if valid
--- a/descartes_core/include/descartes_core/sparse_planner.h
+++ b/descartes_core/include/descartes_core/sparse_planner.h
@@ -1,0 +1,50 @@
+/*
+ * Software License Agreement (Apache License)
+ *
+ * Copyright (c) 2014, Southwest Research Institute
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/*
+ * sparse_planner.h
+ *
+ *  Created on: Dec 17, 2014
+ *      Author: ros developer 
+ */
+
+#ifndef SPARSE_PLANNER_H_
+#define SPARSE_PLANNER_H_
+
+#include <descartes_core/planning_graph.h>
+
+namespace descartes_core
+{
+
+class SparsePlanner : public descartes_core::PlanningGraph
+{
+public:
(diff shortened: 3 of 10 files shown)


descartes_core/CMakeLists.txt
```diff
@@ -65,6 +64,14 @@ if(CATKIN_ENABLE_TESTING)
       test/joint_trajectory_pt.cpp
       test/cartesian_robot.cpp
       test/cartesian_robot_test.cpp)
+      #test/sparse_planner.cpp)
```
shaun-edwards: Did you mean to add the sparse_planner test here or as a separate test?  Is there a reason to have a separate test?

shaun-edwards:
Does this require PR #43 to be accepted first?

descartes_core/CMakeLists.txt
jrgnicho: Yes, I intended to separate that sparse planer test into its own unit test.  Having a separate test is convenient because it runs quicker since it doesn't run all the other test and the terminal output is less confusing to look at.

jrgnicho:
I tried my best to make this PR independent from #43

descartes_core/CMakeLists.txt
shaun-edwards: Can you remove the commented out test?

descartes_core/include/descartes_core/robot_model.h
```diff
@@ -89,6 +89,12 @@ class RobotModel
   virtual bool getFK(const std::vector<double> &joint_pose, Eigen::Affine3d &pose) const = 0;
 
   /**
+   * @brief Returns number of DOFs
+   * @return Int
+   */
+  virtual int getDOF() const = 0;
```
shaun-edwards, line 95: Changes to `RobotModel` will require changes to all classes that implement this interface.  Did you do this?

descartes_core/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 2.8.3)
 project(descartes_core)
 
 set(CMAKE_BUILD_TYPE Debug)
-
+add_definitions(-std=gnu++0x)
```
shaun-edwards, line 5: Are we limiting ourselves be requiring the latest version?  Utilizing _new_ features like the `auto` keyword could present problems.  Do we fully understand what those are?

descartes_core/src/sparse_planner.cpp
```diff
@@ -0,0 +1,562 @@
...
+
+  if(insertGraph(&sparse_trajectory_array) && plan())
+  {
+    int planned_count = sparse_solution_array_.size();
+    int interp_count = cart_points_.size()  - sparse_solution_array_.size();
+    ROS_INFO("Sparse plan succeeded with %i planned point and %i interpolated points",planned_count,interp_count);
+  }
+  else
+  {
+    return false;
```
shaun-edwards, line 68: Log message?  Also, which failed `insertGraph` or `plan`?

descartes_core/src/sparse_planner.cpp
```diff
@@ -0,0 +1,562 @@
...
+  // replanning
+  if(addTrajectory(cp,prev_id,next_id) && plan())
+  {
+    int planned_count = sparse_solution_array_.size();
+    int interp_count = cart_points_.size()  - sparse_solution_array_.size();
+    ROS_INFO("Sparse plan succeeded with %i planned point and %i interpolated points",planned_count,interp_count);
+  }
+  else
+  {
+    return false;
```
shaun-edwards, line 110: Error message?

descartes_core/src/sparse_planner.cpp
```diff
@@ -0,0 +1,562 @@
...
+
+  if(plan())
+  {
+    int planned_count = sparse_solution_array_.size();
+    int interp_count = cart_points_.size()  - sparse_solution_array_.size();
+    ROS_INFO("Sparse plan succeeded with %i planned point and %i interpolated points",planned_count,interp_count);
+  }
+  else
+  {
+    return false;
```
shaun-edwards, line 188: Error message?

descartes_core/src/sparse_planner.cpp
```diff
@@ -0,0 +1,562 @@
...
+  cart_points_[index] = cp;
+  if( plan())
+  {
+    int planned_count = sparse_solution_array_.size();
+    int interp_count = cart_points_.size()  - sparse_solution_array_.size();
+    ROS_INFO("Sparse plan succeeded with %i planned point and %i interpolated points",planned_count,interp_count);
+  }
+  else
+  {
+    return false;
```
shaun-edwards, line 231: Error message?

descartes_core/src/sparse_planner.cpp
```diff
@@ -0,0 +1,562 @@
...
+      return false;
+    }
+
+    sparse_solution_array.push_back(std::make_tuple(index,cp,jp));
+  }
+
+  return true;
+}
+
+bool SparsePlanner::getOrderedSparseCartesianArray(std::vector<TrajectoryPtPtr>& sparse_array)
```
shaun-edwards: There's no guarantee that the array will be a Cartesian array, just that it will be a vector of `TrajectoryPt`'s.  Although in most case it will be, the design of `TrajectoryPt` is such that it doesn't have to.

descartes_core/src/sparse_planner.cpp
```diff
@@ -0,0 +1,562 @@
...
+  }
+
+  return succeeded;
+}
+
+int SparsePlanner::interpolateSparseTrajectory(const SolutionArray& sparse_solution_array,int &sparse_index, int &point_pos)
+{
+  // populating full path
+  joint_points_map_.clear();
+  std::vector<double> start_jpose, end_jpose, rough_interp, aprox_interp, seed_pose(robot_model_->getDOF(),0);
```
shaun-edwards, line 514: I don't see why `getDOF` is required.  Can't it's size be inferred  by the start and end pose dimensions?

descartes_core/include/descartes_core/sparse_planner.h
```diff
@@ -0,0 +1,86 @@
...
+#ifndef SPARSE_PLANNER_H_
+#define SPARSE_PLANNER_H_
+
+#include <descartes_core/planning_graph.h>
+#include <tuple>
+
+namespace descartes_core
+{
+
+class SparsePlanner : public descartes_core::PlanningGraph
```
shaun-edwards, line 34: Should the `SparsePlanner` contain a `PlanningGraph` or inherit from it?  It appears as though you re-implement all of the functionality.  In that case, it really isn't inheritance.

descartes_core/include/descartes_core/sparse_planner.h
```diff
@@ -0,0 +1,86 @@
...
+{
+
+public:
+  typedef std::vector<std::tuple<int,TrajectoryPtPtr,JointTrajectoryPt> > SolutionArray;
+public:
+  SparsePlanner(RobotModelConstPtr &model,double sampling = 0.1f);
+  virtual ~SparsePlanner();
+
+  void setSampling(double sampling);
+  bool setTrajectoryPoints(const std::vector<TrajectoryPtPtr>& traj);
```
shaun-edwards: I'm not sure where these method names come from (if they are inherited), but they could be less wordy.  Specifying TrajectoryPoint is not required since, the argument type implies that.  It could simply be `setPoint` (same for functions below).

descartes_core/include/descartes_core/sparse_planner.h
shaun-edwards, line 34: Doxygen comments.

descartes_core/test/sparse_planner.cpp
```diff
@@ -0,0 +1,78 @@
...
+  for(auto& e:joint_bounds)
+  {
+    double d = (std::get<1>(e)- std::get<0>(e))/NUM_DENSE_POINTS;
+    deltas.push_back(d);
+  }
+
+  // creating trajectory points
+  std::vector<double> joint_vals(deltas.size(),0);
+  traj.reserve(NUM_DENSE_POINTS);
+  for(int i = 0 ; i < NUM_DENSE_POINTS; i++)
```
shaun-edwards, line 59: Should use `++i` instead of `i++` (I forget why).

descartes_core/test/sparse_planner.cpp
```diff
@@ -0,0 +1,78 @@
...
+#include <gtest/gtest.h>
+#include <tuple>
+
+using namespace descartes_core;
+typedef std::vector<descartes_core::TrajectoryPtPtr> Trajectory;
+const int NUM_DENSE_POINTS = 1000;
+Trajectory createTestTrajectory();
+Trajectory TEST_TRAJECTORY = createTestTrajectory();
+
+class TestPoint: public descartes_core::CartTrajectoryPt
```
shaun-edwards, line 14: This should inherit from `TrajectoryPt`.

descartes_core/test/sparse_planner.cpp
```diff
@@ -0,0 +1,78 @@
...
+    joint_pose.assign(vals_.begin(),vals_.end());
+    return true;
+  }
+
+protected:
+
+  std::vector<double> vals_;
+};
+
+Trajectory createTestTrajectory()
```
shaun-edwards, line 42: This should probably be a test fixture, but a method works.

descartes_core/src/sparse_planner.cpp
```diff
@@ -0,0 +1,562 @@
...
+    sparse_trajectory_array.push_back(dense_trajectory_array.back());
+  }
+}
+
+bool SparsePlanner::interpolateJointPose(const std::vector<double>& start,const std::vector<double>& end,
+    double t,std::vector<double>& interp)
+{
+  if(start.size() != end.size() && (t > 1 || t < 0))
+  {
+    return false;
```
shaun-edwards, line 436: Error message.

descartes_core/CMakeLists.txt
jrgnicho: +1

shaun-edwards:
General comments:
- I'm worried that the concept of a Cartesian trajectory is embedded in the planner.  The planner should work on any trajectory point and not make an assumption of a Cartesian point.
- The add/modify/remove point methods seem rather complex to me.  Can they be made simpler?
- How is sampling implemented?  Sampling by time seems right, but time hasn't been integrated into the trajectory points.  Should we just sample by point index?

I've created a [fork](https://github.com/swri-robotics/descartes).  I believe we should accept this and your previous PR to the `amos_integration` branch in order to make forward progress.  I'd still like to address the above items if possible.

PR #47, opened 2015-01-20
Functional Sparse Planner

jrgnicho:
- Functional implementation of the sparse planner
- Completed unimplemented methods of the JointTrajectoryPt class
- Unit test for sparse planner

This PR is meant to capture the current working state of the SparsePlanner class.  However, there are several issues here (mainly related to lack of proper coding practices) that will be address in subsequent PRs.

### DIFF
--- a/descartes_core/include/descartes_core/cart_trajectory_pt.h
+++ b/descartes_core/include/descartes_core/cart_trajectory_pt.h
@@ -70,8 +70,8 @@ struct ToleranceBase
     #param tol Total tolerance zone (assumed symetric about nominal)
     */
   template<typename T>
-  static T createSymmetric(const double x, const double y, const double z,
-                                          const double tol)
+  static T createSymmetric(const double& x, const double& y, const double& z,
+                                          const double& tol)
   {
     return(createSymmetric<T>(x, y, z, tol, tol, tol));
   }
@@ -81,7 +81,7 @@ struct ToleranceBase
     @param x, y, z
     */
   template<typename T>
-  static T zeroTolerance(const double x, const double y, const double z)
+  static T zeroTolerance(const double& x, const double& y, const double& z)
   {
     return(createSymmetric<T>(x, y, z, 0.0, 0.0, 0.0));
   }
@@ -174,8 +174,8 @@ struct TolerancedFrame: public Frame
     orientation_tolerance = ToleranceBase::createSymmetric<OrientationTolerance>(rpy(2),rpy(1),rpy(0),0);
   };
 
-  TolerancedFrame(const Eigen::Affine3d &a, const PositionTolerance pos_tol,
-                  const OrientationTolerance orient_tol) :
+  TolerancedFrame(const Eigen::Affine3d &a, const PositionTolerance &pos_tol,
+                  const OrientationTolerance &orient_tol) :
     Frame(a), position_tolerance(pos_tol), orientation_tolerance(orient_tol){}
 
   PositionTolerance             position_tolerance;
--- a/descartes_core/include/descartes_core/cart_trajectory_pt.h
+++ b/descartes_core/include/descartes_core/cart_trajectory_pt.h
@@ -160,18 +160,18 @@ struct TolerancedFrame: public Frame
   {
     Eigen::Vector3d t = a.translation();
     Eigen::Matrix3d m = a.rotation();
-    Eigen::Vector3d rpy = m.eulerAngles(2,1,0);
+    Eigen::Vector3d rxyz = m.eulerAngles(0,1,2);
     position_tolerance =  ToleranceBase::createSymmetric<PositionTolerance>(t(0),t(1),t(2),0);
-    orientation_tolerance = ToleranceBase::createSymmetric<OrientationTolerance>(rpy(2),rpy(1),rpy(0),0);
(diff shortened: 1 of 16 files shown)


descartes_core/include/descartes_core/cart_trajectory_pt.h
```diff
@@ -70,8 +70,8 @@ struct ToleranceBase
     #param tol Total tolerance zone (assumed symetric about nominal)
     */
   template<typename T>
-  static T createSymmetric(const double x, const double y, const double z,
-                                          const double tol)
+  static T createSymmetric(const double& x, const double& y, const double& z,
```
shaun-edwards, line 73: Do we need to pass these values by reference?  Are they changed by the createSymmetric method?

descartes_core/include/descartes_core/cart_trajectory_pt.h
```diff
@@ -81,7 +81,7 @@ struct ToleranceBase
     @param x, y, z
     */
   template<typename T>
-  static T zeroTolerance(const double x, const double y, const double z)
+  static T zeroTolerance(const double& x, const double& y, const double& z)
```
shaun-edwards, line 84: Same question as above.

descartes_core/include/descartes_core/cart_trajectory_pt.h
```diff
@@ -160,22 +160,22 @@ struct TolerancedFrame: public Frame
   {
     Eigen::Vector3d t = a.translation();
     Eigen::Matrix3d m = a.rotation();
-    Eigen::Vector3d rpy = m.eulerAngles(2,1,0);
+    Eigen::Vector3d rxyz = m.eulerAngles(0,1,2);
```
shaun-edwards, line 163: Can you explain why this was required?  (I just want clarity on this for the future, so that we (probably just me) don't make this mistake again)

jrgnicho, line 163: The axis of rotation corresponding to the roll pitch yaw denominations was not clearly defined in most of the literature that I found on this.  In order to avoid further confusing myself I renamed the variable to something indicative to the axis of rotations and the order in which they are applied.  In this case, the combined rotation Rxyz is the product of R(rxyz(0)) \* R(rxyz(1)) \* R(rxyz(2)) in that specific order.

descartes_core/include/descartes_core/cart_trajectory_pt.h
jrgnicho, line 73: I thought that passing by reference was the original intent since in the original version of this function the 'const' attribute proceeded all the arguments.  When passing by value then 'const' has no effect on the argument as far as I know.

descartes_core/src/cart_trajectory_pt.cpp
```diff
@@ -203,7 +211,78 @@ bool CartTrajectoryPt::getClosestJointPose(const std::vector<double> &seed_state
                                            const RobotModel &model,
                                            std::vector<double> &joint_pose) const
 {
-  NOT_IMPLEMENTED_ERR(false);
+  Eigen::Affine3d nominal_pose,candidate_pose;
```
shaun-edwards, line 214: We should add an issue to address this incomplete method.  I believe it will only work in the simplest of cases.

descartes_core/src/cart_trajectory_pt.cpp
```diff
@@ -203,7 +211,78 @@ bool CartTrajectoryPt::getClosestJointPose(const std::vector<double> &seed_state
...
+
+  if(!model.getFK(seed_state,candidate_pose))
+  {
+    return false;
```
shaun-edwards, line 219: Should we log an error/warning/info/debug message.  It would be nice to know when the interpolation fails.  Since it should fail rarely, we should probably log something.

descartes_core/src/planning_graph.cpp
```diff
@@ -995,9 +997,18 @@ double PlanningGraph::linearWeight(JointTrajectoryPt start, JointTrajectoryPt en
     if (start_vector.size() == end_vector.size())
     {
       double vector_diff = 0;
+      double joint_diff = 0;
       for (int i = 0; i < start_vector.size(); i++)
       {
-        vector_diff += fabs(end_vector[i] - start_vector[i]);
+        joint_diff = std::abs(end_vector[i] - start_vector[i]);
```
shaun-edwards, line 1003: Add an issue to better address large joint changes.  Large joint moves are achievable as long as they happen "far" apart.

shaun-edwards:
Add issues and address (minor) comments.

descartes_core/src/cart_trajectory_pt.cpp
jrgnicho, line 214: Agreed, the current approach may not be sufficient to address the more complex cases.  Issue #43 has been updated to include this observation.

descartes_core/src/cart_trajectory_pt.cpp
jrgnicho, line 219: This kind of error is very unlikely but if it were to happen it'd be much easier to troubleshoot if we had an error message for this.

descartes_core/src/planning_graph.cpp
jrgnicho, line 1003: issue #48 has been created

Issue #46, opened 2015-01-13
descartes robot model as plugin


jrgnicho:
In order to use RobotModel implementations in a generalized way then each Implementation should be made into a plugin that can be loaded at run time.  Currently this won't be possible due to a limitation that prevents libraries containing the plugins from being loaded dynamically if they are also linked statically.  The descartes_core package where the RobotModel interface resides suffers from this drawback since it also contains the implementation for the various trajectory point classes which are linked statically.  One easy solution would be to isolate the RobotModel class into its own package.  The [nav_core](https://github.com/DLu/navigation/tree/hydro/nav_core) of the navigation repo is a good example of this.

jrgnicho:
I propose to divide the descartes packages in the following manner:
- descartes_core: Only interfaces such as:
  a. TrajectoryPt
  b. RobotModel
  c. PlannerInterfaces (not yet defined)
  d. Any future interfaces that may be needed.
- descates_trajectory: Existing TrajectoryPt implementations. These include
  a. CartesiantrajectoryPt
  b. JointTrajectoryPt 
  c. AxialSymetricPt (not in descartes)
  d. Also any future implementations from this interface.
- descartes_planner:  This will require the definition of a new interface for all planners but it should include the these two currently available planners:
  a. PlanningGraph
  b. SparsePlanner
- descartes_robot:
  a. MoveitPlannerAdapter.
  b. Robot specific implementations with their unique close form ik code.

This change will most likely break packages in descartes and the amos repo but in the current state of things this can be easily addressed by replacing namespaces.

Jmeyer1292:
This was implemented a long time ago. See #49.

PR #49, opened 2015-01-23
Issue #46 package rearrangement

jrgnicho:
-The most important change in this PR is that  the class implementations of the various interfaces have been moved into their own packages.  This will partially address issue #46.
- Added initialize(..) method to the RobotModel class in order to use it whenever one of its implementations is instantiated through a plugin.  The proper signature for this method may need some discussion.
- Created a descartes_core::RobotModel plugin from the descartes_moveit::MoveitStateAdaptor class.
- Started the descartes_core::PathPlannerBase interface which will be used to implement current and future path planner classes.  The final form of this interface will need some discussion as well.  Once the planner interface is finalized then the PlannerGraph and SparsePlanner should be modified in order to inherit from this interface.


descartes_core/include/descartes_core/path_planner_base.h
```diff
@@ -0,0 +1,40 @@
...
+#define DESCARTES_CORE_PATH_PLANNER_BASE_H_
+
+#include <descartes_core/trajectory_pt.h>
+#include <descartes_core/robot_model.h>
+#include <vector>
+
+namespace descartes_core
+{
+
+class PathPlannerBase
```
shaun-edwards, line 44: Can we document this class.  Since it is a base/interface class, it should be very clear what these methods do.

descartes_core/include/descartes_core/path_planner_base.h
```diff
@@ -0,0 +1,40 @@
...
+{
+public:
+  virtual ~PathPlannerBase(){}
+
+  virtual bool initialize(RobotModelConstPtr &model,const std::vector<TrajectoryPtPtr>& traj,
+            double sampling = 0.1f) = 0;
+
+  virtual bool addAfter(const TrajectoryPt::ID& ref_id,TrajectoryPtPtr cp) = 0;
```
shaun-edwards: These modifiers should be removed from the base class at some point.  However, we should go through the effort of defining their function since modifiers will be required, albeit in another class.

shaun-edwards:
@jrgnicho, please address comments in planner base class.  Also, do these changes also allow for `TrajectoryPts` to be used as plugins?  This may be desirable in the future.

jrgnicho:
This planner interface operates on TrajectoryPt pointers which is an interface for all the trajectory point specializations.  This setup should allow working with points that were instantiated as plugins.  One current limitation is that there isn't an "init" interface method in TrajectoryPt that each specialization is required to implement.  Defining such method isn't trivial since each type of point's  constructor (Cart, Joint, Axial, etc) takes very different data.

jrgnicho:
Comments have been added to the planner base class.

descartes_core/include/descartes_core/path_planner_base.h
```diff
@@ -0,0 +1,53 @@
...
+
+class PathPlannerBase
+{
+public:
+  virtual ~PathPlannerBase(){}
+
+  /**
+   * @brief Plans a path for the given robot model through the points in the input trajectory.
+   * @param model robot model implementation for which to plan a path
+   * @param sampling This value will be interpreted differently by each planner specialization.
```
shaun-edwards: While it can be interpreted differently, it must have a similar use.  We must be more specific here.

descartes_core/include/descartes_core/path_planner_base.h
```diff
@@ -0,0 +1,53 @@
...
+   */
+  virtual bool initialize(RobotModelConstPtr &model, double sampling = 0.1f) = 0;
+
+  /**
+   * @brief Generates a robot path from the trajectory.
+   * @param traj the points used to plan the robot path
```
shaun-edwards, line 70: Should we be more specific here?  Doesn't `planPath` only ever return a `JointTrajectoryPt`?  Is there a benefit in returning the more generic base class?

shaun-edwards, line 70: I think we need a more informative `planPath` method.  When a planner does fail, it would be nice to know why.  Simply checking of IK solution is not enough since the planning graph can still have disconnects.  If `planPath` returns `false`, mabe traj could be populated with the path segment that could be planned (i.e from the start to the disconnect).

shaun-edwards, line 70: Oops...I guess these comments really apply to the `getPath` method.

descartes_core/include/descartes_core/path_planner_base.h
```diff
@@ -0,0 +1,53 @@
...
+   */
+  virtual bool planPath(const std::vector<TrajectoryPtPtr>& traj) = 0;
+
+  /**
+   * @brief Returns the robot path generated from the input trajectory
+   * @param path Array of path points
+   */
+  virtual bool getPath(std::vector<TrajectoryPtPtr>& path) = 0;
```
shaun-edwards, line 79: Why is this needed?  Couldn't `planPath` return this?

descartes_core/include/descartes_core/path_planner_base.h
```diff
@@ -0,0 +1,53 @@
...
+
+  virtual bool addAfter(const TrajectoryPt::ID& ref_id,TrajectoryPtPtr tp) = 0;
```
shaun-edwards, line 85: Could we simply have one `addBefore/After` method...i.e `add`.  The behavior could always be add _after_.  If someone want to add a first point, then a `ref_id` of `nil` could be set.

descartes_core/include/descartes_core/path_planner_base.h
jrgnicho, line 70: One reason that 'getPath(...)' returns an array of generic TrajectoryPts is because the JointTrajectoryPt class has been moved out of descartes_core into the descartes_trajectory package.  
Furthermore, I agree that in most cases only joint values (JointTrajectoryPts) need to be returned, however future planner implementations may need to return additional data ( such as tool process data) or even point transition flags (linear,  joint move, tolerance zones) which could be applicable when translating these path points to robot programming language commands.  Thus, using the generic point type will allow returning point specialization for such cases.

descartes_core/include/descartes_core/path_planner_base.h
jrgnicho, line 79: This is convenient to have whenever one needs to get the last valid path without having to replan it

descartes_core/include/descartes_core/path_planner_base.h
jrgnicho, line 85: Yes, I'm on board with that

descartes_core/include/descartes_core/path_planner_base.h
jrgnicho: Since the initialization parameters will likely vary significantly for each planner implementation then would it be better to pass a map<string,string> configuration argument that will be parsed by each planner?

descartes_core/include/descartes_core/path_planner_base.h
jrgnicho, line 70: Regarding failure handling, perhaps the correct way is to return an error code from an enumeration.  We might also want to create an 'PlannerError' structure which contains the error code, the trajectory point where the error occurred and a string with a brief error description.

descartes_core/include/descartes_core/path_planner_base.h
shaun-edwards: I seem to remember OMPL being done this way...sounds like it will work.

descartes_core/include/descartes_core/path_planner_base.h
shaun-edwards, line 70: All good points....can you think of examples of different failures and how they error structure would convey that?  Error strings are not useful programatically, although they could be helpful on a GUI.

jrgnicho, line 70: The way that I envisioned using such error structure in the planPath(...) method would be as follows:

```
int planPath(const std::vector<TrajectoryPt>& traj, PlannerError& error_struct)
```

where the returned int would be an error code defined in an enumeration and the error_struct  would look as follows:

```
struct PlannerError
{
  std::string error_message;
  int error_code;
  TrajectoryPt::ID id
}
```

In any case, the error codes would have to be very generic, the ones that come to mind are unavailable IK, high joint velocities, self collision, collision with external objects, planning time exceeded, incorrect DOFs, mismatching link or tool names.
The error strings could be very helpful when troubleshooting issues with a specific planner implementation.

gavanderhoorn, line 70: Perhaps it would be enough to provide the `error_code` member, and then have the caller use an additional function to map it to an error string? Analogous to how `strerror()` and `errno` work?

gavanderhoorn, line 70: Also: `PlanningResult` (or something similar) seems more generic?

descartes_core/include/descartes_core/path_planner_base.h
gavanderhoorn, line 85: On the other hand, a method called `addAfter(..)` immediately tells me what it is going to do (less confusing, easier to read).

Perhaps do what Shaun suggests, but keep `addAfter(..)` and just have it call `add(..)` with `ref_od=nil`?

descartes_core/include/descartes_core/path_planner_base.h
jrgnicho, line 70: PlanningResult sounds more appropriate to me since it won't be an error when it works.

descartes_core/include/descartes_core/path_planner_base.h
jrgnicho, line 85: That's a good point that @gavanderhoorn  brought up in regards to clarity.  I think we should reserve the nil ID value for checking other erroneous conditions.  Also, having both addAfter/Before methods will make it easier to modify the path since the calling code doesn't need to retrieve the preceding id when the desired action is to add before.

shaun-edwards:
@jrgnicho...based on our discussion...can you make any changes and let me know when things are ready to merge?

jrgnicho:
The commits 3306597 to 2f69bb1 contain the following changes:
- Completed definition of the PathPlannerBase interface as discussed above.
- Wrap the PlanningGraph class into the DensePlanner class which implements PathPlannerBase.
- The SparsePlanner class now implements PathPlannerBase as well.
- Minor improvements were made to the PlanningGraph in order to prevent memory leaks when copying into  runaway pointers.
- Added/Updated ros unit test for the Sparse and Dense planners.

The unit test can be run from the catkin workspace directory as follows:

```
catkin_make run_tests_descartes_planner
```

There are still some that should probably be addressed in future PR's
- Error messages returned by the getErrorMessage(...) interface method are pretty generic and may not convey enough information that lead to a planning failure.
- Collision checking still needs to be implemented (issue #50)
- Interpolator planner class also needs to be implemented (issue #51).

I will create issues for those that don't have one yet.  @shaun-edwards  please review.

shaun-edwards:
+1

Issue #50, opened 2015-02-06
Robot Model Collision Detection


jrgnicho:
The MoveitStateAdapter does not perform any collision checking. This capability needs to be implemented so that only IK and FK collision free solutions are returned.

jrgnicho:
Addressed in #50

PR #54, opened 2015-02-10
Issue #50 robot model collision

jrgnicho:
This PR should be reviewed after #49 is accepted.
- It addresses the issue described in #50 regarding collision checks.
- In addition to that, it added two interface methods for the descartes_core::RobotModel  for setting and getting a map containing option key/value pairs.  
- The constructor for the MoveitStateAdapter no longer takes any initialization arguments, instead it takes them in the 'initialize(..)' interface method.
- The unit test have been updated to conform to the changes above.

### DIFF
--- a/descartes_core/include/descartes_core/robot_model.h
+++ b/descartes_core/include/descartes_core/robot_model.h
@@ -26,6 +26,8 @@
 namespace descartes_core
 {
 
+typedef std::map<std::string,std::string> RobotModelOptions;
+
 DESCARTES_CLASS_FORWARD(RobotModel);
 
 /**@brief RobotModel defines the interface to a kinematics/dynamics functions.  Implementations
@@ -101,6 +103,16 @@ public:
    */
   virtual bool initialize(const std::string robot_description, const std::string& group_name,
                           const std::string& world_frame,const std::string& tcp_frame) = 0;
+
+  /* @brief sets the configuration options.
+   * @param options map containing the key/value option pairs.
+   */
+  virtual bool setOptions(const RobotModelOptions& options) = 0;
+
+  /* @brief returns the current options used by this robot model
+   * @return RobotModelOptions map
+   */
+  virtual RobotModelOptions getOptions() = 0;
 
 protected:
 
--- a/descartes_core/include/descartes_core/robot_model.h
+++ b/descartes_core/include/descartes_core/robot_model.h
@@ -26,8 +26,6 @@
 namespace descartes_core
 {
 
-typedef std::map<std::string,std::string> RobotModelOptions;
-
 DESCARTES_CLASS_FORWARD(RobotModel);
 
 /**@brief RobotModel defines the interface to a kinematics/dynamics functions.  Implementations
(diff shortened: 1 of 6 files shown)


shaun-edwards:
@jrgnicho, can you fix this PR.  The diff shows changes from already accepted PRs.

jrgnicho:
OK, will do

jrgnicho:
@shaun-edwards , the issue with the old diffs has been fixed and some minor changes for setting an options map into the MoveitStateAdapter have been made as well.  
Please review

jrgnicho:
@shaun-edwards or @Jmeyer1292  ping

jrgnicho:
@shaun-edwards I added setter and getter methods to the RobotModel interface for enabling and disabling collision checks and removed the options map.  Please review

descartes_moveit/include/descartes_moveit/moveit_state_adapter.h
```diff
@@ -85,7 +86,16 @@ class MoveitStateAdapter : public descartes_core::RobotModel
    * @brief Pointer to moveit robot state (mutable object state is reset with
    * each function call
    */
+
+  /**
+   * TODO: Checks for collisions at this joint pose. The setCollisionCheck(true) must have been
```
shaun-edwards, line 91: This seems counter-intuitive.  I can see using the flag to turn on/off collision checking for IK, but anybody calling this method should expect it to perform collision checking.
<EDIT>
Should we consider adding this method to the base class?

descartes_core/include/descartes_core/robot_model.h
```diff
@@ -102,9 +102,31 @@ class RobotModel
...
+  virtual bool getCheckCollisions()
+  {
+    return check_collisions_;
+  }
+
+
 protected:
 
-  RobotModel(){}
+  RobotModel(): check_collisions_(false){}
```
shaun-edwards, line 127: Should the default collision checking be `true`?  What do you think?

shaun-edwards:
Should we consider using the `isValid` base methods to check collisions as well.  This would require ensuring that their current use still makes sense.

shaun-edwards:
Couple of comments/items for discussion, but in general I think we should move forward.

descartes_moveit/include/descartes_moveit/moveit_state_adapter.h
jrgnicho, line 91: This is a protected method so it is not meant to be called directly by the implementer.  I placed the collision code inside this method so that it can be conveniently invoked when needed by any of the interface methods.  There is already an **isValid** method which should also check for collisions when these are enabled; adding this method to the interface might make matters more confusing in my opinion.

descartes_core/include/descartes_core/robot_model.h
jrgnicho, line 127: Since collision checks are expensive operations I think they should be disabled by default.

jrgnicho:
The **isValid** method of the MoveitStateAdaptor in this PR checks for collisions when these are enabled.

jrgnicho:
@shaun-edwards  if there aren't any more issues to address here then this PR can be merged.

shaun-edwards:
+1

PR #56, opened 2015-02-15
Parallel IK Calculations

Jmeyer1292:
This pull request adds a clone function to the RobotState interface and uses this to do parallel IK calculations in the calculateJointSolutions function.

This is not necessarily the best solution. But it was the quickest and cleanest solution that I could imagine, and I wanted it for my own work. I tested this on an Indigo machine with gcc 4.8 and it makes use of c++11 features (since we were using them already).

### DIFF
--- a/descartes_core/src/planning_graph.cpp
+++ b/descartes_core/src/planning_graph.cpp
@@ -45,9 +45,9 @@ namespace
   using ConstTrajectorIterator = std::vector<descartes_core::TrajectoryPtPtr>::const_iterator;
 
   VecJointSolutions
-  calculateJointSolutions(descartes_core::RobotModelPtr model, 
-                          ConstTrajectorIterator begin,
-                          ConstTrajectorIterator end)
+  rangeCalculateJointSolutions(descartes_core::RobotModelPtr model, 
+                               ConstTrajectorIterator begin,
+                               ConstTrajectorIterator end)
   {
     VecJointSolutions solutions;
     solutions.reserve(std::distance(begin, end));
@@ -91,7 +91,7 @@ namespace
       // Create a seperate copy of the robot state for this worker
       descartes_core::RobotModelPtr state_copy = model.clone();
       // Launch task
-      futures[i] = std::async(std::launch::async, calculateJointSolutions, state_copy, start, end);
+      futures[i] = std::async(std::launch::async, rangeCalculateJointSolutions, state_copy, start, end);
     }
 
     // Wait for solutions
--- a/descartes_core/src/planning_graph.cpp
+++ b/descartes_core/src/planning_graph.cpp
@@ -42,22 +42,32 @@ namespace
 {
   using JointSolutions = std::vector<std::vector<double>>;
   using VecJointSolutions = std::vector<JointSolutions>;
-  using ConstTrajectorIterator = std::vector<descartes_core::TrajectoryPtPtr>::const_iterator;
+  using ConstTrajectoryIterator = std::vector<descartes_core::TrajectoryPtPtr>::const_iterator;
 
-  VecJointSolutions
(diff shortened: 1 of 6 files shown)


gavanderhoorn:
> [..] and it makes use of c++11 features (since we were using them already).

I can't remember where exactly, but I think @shaun-edwards commented on the use of C++11 in another consortium project some time ago. I'm fine with using it, but it does mean that people on Hydro/Precise won't be able to build this (without installing newer compilers) and it cannot be released for Hydro on the build farm.

This being an issue depends on the requirements set up for descartes, obviously. If Hydro/Precise is not one of the targeted/supported platforms, then there is no issue in merging (apart from any potential functional issues).

I like the idea of this PR though. A possible alternative would be to use Boost threads, etc?

gavanderhoorn:
> [..] I'm fine with using it, but it does mean that people on Hydro/Precise won't be able to build this (without installing newer compilers) and it cannot be released for Hydro on the build farm.

Ok, so according to [C++0x/C++11 Support in GCC](https://gcc.gnu.org/projects/cxx0x.html) that might not be entirely true ([GCC-4.6](https://gcc.gnu.org/gcc-4.6/cxx0x_status.html) vs [GCC-4.8](https://gcc.gnu.org/gcc-4.8/cxx0x_status.html)).

I do think however that it would be good to clear this up. Maybe extend the readme with a section on the targeted/supported platforms?

jrgnicho:
There is currently a PR #49 which makes some major changes to the directory structure of the entire repo.  Furthermore, there are also some important interface changes/additions.  The changes in this PR may not play nicely with those in #49.  I think the best way to deal with the potential future conflicts is to merge #49 into hydro-devel first and then merge this PR manually if that is needed.

descartes_core/src/planning_graph.cpp
```diff
@@ -805,36 +878,32 @@ bool PlanningGraph::calculateJointSolutions()
...
-    ROS_INFO("CartID: %s: JointPoses count:%i", boost::uuids::to_string(tempID).c_str(), (int)(joint_poses.size()));
+  for (const auto& kv : *cartesian_point_link_) // could also use lambda and std::for_each
+  {
+    trajectory.push_back(kv.second.source_trajectory_);
+  }
+  
+  // Compute joint solutions for trajectory pts
+  VecJointSolutions all_joint_solutions = parallelCalculateJointSolutions(trajectory, *robot_model_, max_threads_);
 
-    if (joint_poses.size() == 0)
```
jrgnicho, line 821: This error message should probable be kept since it is useful for troubleshooting.

descartes_core/src/planning_graph.cpp
```diff
@@ -29,24 +29,97 @@
...
+  {
+    VecJointSolutions solutions;
+    solutions.reserve(std::distance(begin, end));
+
+    while (begin != end)
+    {
+      JointSolutions joint_poses;
+      begin->get()->getJointPoses(*model, joint_poses);
+
+      solutions.push_back(std::move(joint_poses));
```
jrgnicho: A set of joint solutions may not always be available.  Perhaps the appropriate way to handle this would be to quit out of the async  process and print an error message as it was done in the single threaded version.

descartes_moveit/src/moveit_state_adapter.cpp
```diff
@@ -86,7 +86,7 @@ bool MoveitStateAdapter::getIK(const Eigen::Affine3d &pose, std::vector<double>
 
 
   if (robot_state_->setFromIK(robot_state_->getJointModelGroup(group_name_), tool_pose,
-                              tool_frame_))
+                              tool_frame_, 1, 0.05))
```
jrgnicho: This values should be placed into global constants.

Jmeyer1292:
I'm closing this pull request and will open a new one that takes into account the new architecture.

PR #58, opened 2015-02-19
Added functions for running IK solutions in multiple threads

Jmeyer1292:
This is a repeat of pull-request #56. This adds a clone function to the RobotModel interface and spawns a given number of threads each with its own clone of the RobotModel to do IK calculations in parallel. 

By default one thread is used, so multi-threading is opt-in, though new thread and clone are made and the original blocks till the former is done.

### DIFF
--- a/descartes_planner/src/planning_graph.cpp
+++ b/descartes_planner/src/planning_graph.cpp
@@ -36,16 +36,108 @@
 #include <boost/uuid/uuid_io.hpp> // streaming operators
 #include <boost/graph/dijkstra_shortest_paths.hpp>
 
+#include <future>
+
 using namespace descartes_core;
 using namespace descartes_trajectory;
+
+namespace
+{
+  using JointSolutions = std::vector<std::vector<double>>;
+  using VecJointSolutions = std::vector<JointSolutions>;
+  using ConstTrajectoryIterator = std::vector<descartes_core::TrajectoryPtPtr>::const_iterator;
+
+  using JointSolutionResult = std::pair<VecJointSolutions, bool>; // result and flag if everything went well
+
+  JointSolutionResult
+  rangeCalculateJointSolutions(descartes_core::RobotModelPtr model, 
+                               ConstTrajectoryIterator begin,
+                               ConstTrajectoryIterator end)
+  {
+    JointSolutionResult solutions;
+    solutions.first.reserve(std::distance(begin, end));
+    solutions.second = true;
+
+    while (begin != end)
+    {
+      JointSolutions joint_poses;
+      begin->get()->getJointPoses(*model, joint_poses);
+
+      if (joint_poses.empty())
+      {
+        solutions.second = false;
+        ROS_ERROR("No joint solutions available for point %s", 
+          boost::uuids::to_string(begin->get()->getID()).c_str());
+      }
+
+      solutions.first.push_back(std::move(joint_poses));
+      ++begin;
+    }
+
+    return solutions;
+  }
+
+  JointSolutionResult
+  parallelCalculateJointSolutions(const std::vector<descartes_core::TrajectoryPtPtr>& trajectory,
+                                  const descartes_core::RobotModel& model,
+                                  const unsigned max_threads)
+  {
+    // Might want to insert check to see if work to be done is small and if so,
+    // just run the IKs synchronously
+    JointSolutionResult solutions;
+    solutions.first.reserve(trajectory.size());
+    solutions.second = true; // success by default
+
+    // Divide work into chunks
+    const unsigned chunk_size = trajectory.size() / max_threads;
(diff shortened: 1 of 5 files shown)


shaun-edwards:
@jrgnicho, please review.

Has this change passed the path planning unit test?

descartes_planner/src/planning_graph.cpp
```diff
@@ -36,16 +36,108 @@
...
+    while (begin != end)
+    {
+      JointSolutions joint_poses;
+      begin->get()->getJointPoses(*model, joint_poses);
+
+      if (joint_poses.empty())
+      {
+        solutions.second = false;
+        ROS_ERROR("No joint solutions available for point %s", 
+          boost::uuids::to_string(begin->get()->getID()).c_str());
```
jrgnicho: This isn't necessarily an issue but in order to avoid doing all these conversions into a c_string you could use the ROS_???_STREAM variant for printing uuids since the uuid_io header already provides an overloaded insertion operator that works with those.

descartes_planner/src/planning_graph.cpp
```diff
@@ -36,16 +36,108 @@
...
+
+    // Wait for solutions
+    for (auto& future : futures)
+    {
+      JointSolutionResult solution_segment = std::move(future.get());
+
+      solutions.second = solutions.second && solution_segment.second; // carry the false flag forward if is returned
+      if (!solution_segment.second)
+      {
+        ROS_ERROR("One segment of the calculation failed to find a solution for all points");
```
jrgnicho: Should this error condition exit the for loop and/or the function?

descartes_planner/src/planning_graph.cpp
Jmeyer1292: I could. I was just trying to keep with the style of the existing code.

jrgnicho:
In the newer descartes architecture the user interacts directly with either the DensePlanner or SparsePlanner which in turn use the PlanningGraph internally.  As things stand, the PlanningGraph is inaccessible and  will default to a single threaded mode since it can not be configured differently.  
One way we could get around this is by using the "setConfig(config_map)" method of the planner interface and pass the multithreaded parameters in the map.

descartes_planner/src/planning_graph.cpp
Jmeyer1292: If we were to terminate planning on a bad point, then the signal would need to come from inside a worker thread and then be propagated back to everyone else.

This raises an important design decision though: is a bad point the end of a plan? Or is it preferable to keep calculating and somehow return a list of bad points? Right now the planner warns and continues.

Jmeyer1292:
@jrgnicho That sounds very reasonable.

descartes_planner/src/planning_graph.cpp
Jmeyer1292: Pinging @shaun-edwards for his opinion.

descartes_planner/src/planning_graph.cpp
shaun-edwards: +1 for `ROS_*_STREAM` methods

descartes_planner/src/planning_graph.cpp
shaun-edwards: A bad point (or one without any IK solutions) results in (by definition) no plan.  The current MoveIt cartesian planner ignores bad points and returns a partial path.  We do NOT want to do this.  @jrgnicho and I had some discussion (in person) on this for PR #49.  If I am remembering correctly, we decided that these were the requirements for a planner (@jrgnicho feel free to correct me).
- A planner should fail quickly if plan is not possible.  A proper return code should indicate why.
- Additional, possibly slow methods should allow the user to get more information about the failure.

This approach balances responsiveness with user feedback.  One could imagine an application where knowing whether a plan is possible is all that is required.  Taking additional time to disect the failure, by default, is of no value to the user. However, another application such as a GUI, would want more detailed information about a failure so that the user could correct the problem points.

jrgnicho: In most cases, it'd be preferable to have the planner return as soon as a planning error is encountered. 
For those more select cases where detailed failure information is needed, we could configure the planner to operate in a "debugging/troubleshooting" mode.  This could be accomplished by passing one or more configuration parameters/flags that would be used to set such a mode.

Handling the second more involved case is perhaps outside of the scope of this PR, so it should probably be addressed separately. For now, I suggest that we make the planner quit immediately when a failure is encountered.

Jmeyer1292:
Upon failing to calculate any IK solutions for a given point, all threads will now terminate within an iteration or two and a false value will be correctly returned from calculateJointSolutions.

shaun-edwards:
Closing this request for now.  Other PRs have increased performance to where this may not be required and makes things more complex as a result.  We should consider this in the future should performance become an issue again.

Issue #59, opened 2015-02-19
Const Correctness in the Planner Base Class


Jmeyer1292:
I propose that all of the get\* methods in "path_planner_base.h" be marked const. I think it is reasonable for users to expect that retrieving a plan not change the state of the graph for subsiquent requests and be able to use const correctness in their own code.

Specifically, the following methods should be changed:
virtual void getConfig(PlannerConfig& config) const = 0;
  virtual bool getPath(std::vector<TrajectoryPtPtr>& path) const = 0;
  virtual int getErrorCode() const = 0;
  virtual bool getErrorMessage(int error_code, std::string& msg) const = 0;

Jmeyer1292:
Merged #60. Closing.

PR #60, opened 2015-02-19
Issue59 const correctness

Jmeyer1292:
Per issue #59, this pull request makes changes to the path planner base request to use const qualifiers.

### DIFF
--- a/descartes_planner/include/descartes_planner/dense_planner.h
+++ b/descartes_planner/include/descartes_planner/dense_planner.h
@@ -23,15 +23,15 @@ public:
 
   virtual bool initialize(descartes_core::RobotModelConstPtr &model);
   virtual bool setConfig(const descartes_core::PlannerConfig& config);
-  virtual void getConfig(descartes_core::PlannerConfig& config);
+  virtual void getConfig(descartes_core::PlannerConfig& config) const;
   virtual bool planPath(const std::vector<descartes_core::TrajectoryPtPtr>& traj);
-  virtual bool getPath(std::vector<descartes_core::TrajectoryPtPtr>& path);
+  virtual bool getPath(std::vector<descartes_core::TrajectoryPtPtr>& path) const;
   virtual bool addAfter(const descartes_core::TrajectoryPt::ID& ref_id, descartes_core::TrajectoryPtPtr tp);
   virtual bool addBefore(const descartes_core::TrajectoryPt::ID& ref_id, descartes_core::TrajectoryPtPtr tp);
   virtual bool remove(const descartes_core::TrajectoryPt::ID& ref_id);
   virtual bool modify(const descartes_core::TrajectoryPt::ID& ref_id, descartes_core::TrajectoryPtPtr tp);
-  virtual int getErrorCode();
-  virtual bool getErrorMessage(int error_code, std::string& msg);
+  virtual int getErrorCode() const;
+  virtual bool getErrorMessage(int error_code, std::string& msg) const;
 
 protected:
 
--- a/descartes_planner/include/descartes_planner/sparse_planner.h
+++ b/descartes_planner/include/descartes_planner/sparse_planner.h
@@ -88,7 +88,7 @@ protected:
   };
 
   double sampling_;
-  int error_code_;
+  mutable int error_code_;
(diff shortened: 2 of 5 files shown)


descartes_planner/include/descartes_planner/sparse_planner.h
```diff
@@ -88,7 +88,7 @@ class SparsePlanner: public descartes_core::PathPlannerBase
   };
 
   double sampling_;
-  int error_code_;
+  mutable int error_code_;
```
shaun-edwards: Generally `mutable` is used to get around code that isn't const correct.  Why is it required here?

descartes_planner/include/descartes_planner/dense_planner.h
```diff
@@ -23,15 +23,15 @@ class DensePlanner: public descartes_core::PathPlannerBase
 
   virtual bool initialize(descartes_core::RobotModelConstPtr &model);
   virtual bool setConfig(const descartes_core::PlannerConfig& config);
-  virtual void getConfig(descartes_core::PlannerConfig& config);
+  virtual void getConfig(descartes_core::PlannerConfig& config) const;
   virtual bool planPath(const std::vector<descartes_core::TrajectoryPtPtr>& traj);
```
jrgnicho, line 27: Omitting 'planPath' from being a const accessible method would make the class unusable

descartes_planner/src/sparse_planner.cpp
```diff
@@ -526,23 +526,25 @@ bool SparsePlanner::getPath(std::vector<TrajectoryPtPtr>& path)
...
-int SparsePlanner::getErrorCode()
+int SparsePlanner::getErrorCode() const
 {
   return error_code_;
 }
 
-bool SparsePlanner::getErrorMessage(int error_code, std::string& msg)
+bool SparsePlanner::getErrorMessage(int error_code, std::string& msg) const
 {
-  if(error_map_.count(error_code)>0)
```
jrgnicho, line 543: Just out of curiosity, was there anything incorrect about the previous approach?

descartes_planner/include/descartes_planner/sparse_planner.h
Jmeyer1292: I threw mutable in here because the getPath function of this class checks some conditions and maybe writes to error_code_. I thought it wasn't too much of a violation of the semantics of const-correctness to allow the class to mutate the error-code_ variable.

descartes_planner/include/descartes_planner/dense_planner.h
Jmeyer1292, line 27: You could still use it in a non const-qualified way. planPath is a big, expensive call and mutates basically all of the state in the object. Calling it changes the fundamental nature of the planner.

descartes_planner/src/sparse_planner.cpp
Jmeyer1292, line 543: Logically probably not, but there's no const overload for operator[] in std::map so it won't compile if that function is marked const.

descartes_planner/include/descartes_planner/dense_planner.h
jrgnicho, line 27: +1

descartes_planner/src/sparse_planner.cpp
jrgnicho, line 543: +1

jrgnicho:
+1 for merge

jrgnicho:
@shaun-edwards, would you have any additional remarks?

Jmeyer1292:
@jrgnicho Hold on, I want to look more into the variable I marked mutable and see if there's a prettier solution.

jrgnicho:
OK

Jmeyer1292:
So when SparsePlanner's getPath is called, it checks to see if the solution is empty and internally sets its state to PlannerError::EMPTY_PATH if that is so. Dense planner does not do the same. Is this desired behavior?

Jmeyer1292:
@jrgnicho Please take a look at the latest changes.

jrgnicho:
+1 for merge

jrgnicho:
+1 for merge

PR #61, opened 2015-02-20
General Code Cleanups in Planning Graph

Jmeyer1292:
This pull request does the following:
1. Changes debug statements to *_STREAM type safe varieties and thus removes a bunch of compiler warnings
2. Makes small changes to edge calculate function to reduce copies
3. Cleans up most of the dynamic memory allocation and should fix most of the outstanding leaks.


Jmeyer1292:
@jrgnicho I removed the uses you pointed out. There are still a couple stringstreams left in the code that are used to build a printout of the whole map.

jrgnicho:
Feel free to remove those extra ones that are not collecting information from consecutive steps (for loops, multiple method calls, etc).
+1 for merge after that.

jrgnicho:
+1 for merge

Issue #63, opened 2015-03-07
Graph Pruning, Djikstra's Search Problems, and Related Issues


Jmeyer1292:
TLDR: I discovered issues with how pruning and Djikstras_shortest_paths was being used. Optimal path planning, NOT IK calculations, (using aggressive joint delta checks) has speed up by 2000-5000x in my limited testing.

I have been working on making Descartes lately in my work and made some improvements. I discovered a few issues and have developed a fixes for them. I'm making this issue to reference in my pull-request (coming this weekend) and to share with you all the exciting results. This is related to issue #21 and issue #48.

The two most expensive components of Descartes are the IK solutions and the graph searching to find an ideal solutions. IK is O(n). Searching is something bigger like O(n^2). With IKFast, the IK can be done very quickly but it was taking minutes to solve for the optimal path with several hundred points.

The long planning time can be abridged by fixing the start and end joint positions, but you have to know those positions and it sacrifices optimality.

The reason the planner is so slow right now is primarily two fold:
1. There are checks for large joint changes and on that condition edges are given huge weights when in fact they should simply be removed. Not adding an edge led to strange problems where the graph search would solve paths that were only several points long. This turned out to be an issue with the findStartVertices and findEndVertices which were actually checking all of the vertices for the absence of in-edges and out-edges respectively. Naturally if we start pruning edges then outliers in the middle of the users trajectory will start showing up as start or end points using that logic. Fixing that allows Dijkstra's search to search MUCH faster and still fail elegantly.
2. The boost algorithm being used (http://www.boost.org/doc/libs/1_57_0/libs/graph/doc/dijkstra_shortest_paths.html) calculates the shortest path from a source vertex to EVERY other vertex. Currently we are running the algorithm for every start point to every end point (n^2 ops) when all we need to do is run the algorithm from every start point (n operations). Fixing this obviously means we get N times speed boosts where N is the number of end point discretizations.

Using C++11's steady clock from the chrono library, I took some poor man's measurements using the simulated objects in Godel (the smallest face closest to the robot and the biggest face) and got the following timing (all builds were with release flags):

Time  w/o Pruning
6092 joint solutions -> 10467 ms
39948 joint solutions -> 211514 ms

Time  w/ Pruning
6092 joint solutions -> 37 ms
39948 joint solutions -> 561 ms

Time  w/ Pruning and Boost improvement
6084 joint solutions -> 5 ms
39948 joint solutions -> 41 ms

[...]

PR #64, opened 2015-03-07
Graph Pruning and Dijkstras search adjustments

Jmeyer1292:
See issue #63 for details. This pull-request adds graph pruning to Descartes planning_graph.cpp and also reduces calls to Dijkstra's.

### DIFF
--- a/descartes_planner/src/planning_graph.cpp
+++ b/descartes_planner/src/planning_graph.cpp
@@ -41,7 +41,7 @@ using namespace descartes_trajectory;
 namespace descartes_planner
 {
 
-const double MAX_JOINT_DIFF = M_PI;
+const double MAX_JOINT_DIFF = M_PI / 2;
 const double MAX_EXCEEDED_PENALTY = 10000.0f;
 
 PlanningGraph::PlanningGraph(RobotModelConstPtr model)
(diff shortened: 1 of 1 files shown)


descartes_planner/src/planning_graph.cpp
```diff
@@ -41,7 +41,7 @@ using namespace descartes_trajectory;
 namespace descartes_planner
 {
 
-const double MAX_JOINT_DIFF = M_PI;
+const double MAX_JOINT_DIFF = M_PI / 20;
```
ClayFlannigan: Hard coding this limit scares me.  Physically, it should be a joint velocity limit defined by the robot.  Thus it should be dependent on time and the configuration of the robot.  Is there anyway to expose this at a higher level?

Jmeyer1292: @ClayFlannigan Absolutely. The other half of the work I did was toward what you speak of. The user associates each point with a time window that indicates the time for the robot to realize this point relative to its previous position. 

I added a function to descarte's RobotModel interface that takes as arguments a pair of joint positions (source and target) and a time value and it returns true if the motion is possible in the time stated. This is used to do pruning instead of the existing hard-coded solution.

I put this code up seperately because it is pretty short and sweet. The timing stuff will probably take some iteration before everyone is happy.

shaun-edwards:
@Jmeyer1292, How did you test the results of this change?

Jmeyer1292:
I did not do any formal testing, but I have used it extensively in simulation and in testing actual hardware (for the Godel project). 

I see this as more of a bug fix than a feature addition. 
1. It makes the starting/ending vertices the user's starting and ending vertices and not any vertices without in/out edges. 
2. It also fixes a bug where the calculate edges function will always return true if there's at least one edge in the entire graph. 
3. It uses the boost dijkstras function as intended: `Dijkstra's algorithm finds all the shortest paths from the source vertex to every other vertex by iteratively ``growing'' the set of vertices S to which it knows the shortest path` so I only call it once per starting point instead of once per starting/ending pair.

shaun-edwards:
+1...thanks for the contribution.

Issue #48, opened 2015-01-23
Improvement to handle large joint changes


jrgnicho:
Paths returned by the graph planner sometimes contain large joint changes between consecutive points

shaun-edwards:
The planners should check if the joint changes are possible.  That is whether the robot is physically capable of making such a jump.  Since we are planning for trajectory points (i.e. points with timing data), this should be possible.  Unfortunately, it is more complex than it appears.  Whether a point is physically reachable depends upon the joint velocities at the current point (which is not known to the planner until after the path is planned).  

A possible compromise solution is to eliminate large changes that would require greater than the max velocities.  This essentially ignores acceleration constrains.

A proper solution would require sampling waypoint trajectories as part of the graph, in addition to current tolerances.

Jmeyer1292:
I have found some limited success in changing the way sampling works. Instead of calling robot_state's getRandomPositions() function, I call getRandomPositions(jointgroup*, RandomNumberGenerator).

Right now I'm creating a new random generator for every call to getAllIK and seeding it with a fixed value. It's not efficient or gauranteed to work, but I find it produces much better paths for what I'm working with. 

http://docs.ros.org/indigo/api/random_numbers/html/classrandom__numbers_1_1RandomNumberGenerator.html

Jmeyer1292:
The timing constraints nominally take care of this. Now practically there are other concerns. The other, more subtle, issues should be discussed in a separate PR.

Issue #52, opened 2015-02-06
Moveit GetCartesianPath Service


jrgnicho:
Create a node that services the [moveit_msgs::GetCartesianPath](http://docs.ros.org/api/moveit_msgs/html/srv/GetCartesianPath.html) and uses a descartes path planner

madratman:
I'm new here. Here's an attempt.  
https://github.com/madratman/descartes/blob/issue_%2352_moveit_cartesian_path_service/descartes_moveit/src/cartesian_path_service_trial.cpp
Sorry, I didn't see there was a separate branch for the issue already before working on it and hence I haven't used a class. And came to know later that there is a  descartes/descartes_trajectory/src/cartesian_interpolator.cpp as well.  

I have translated from moveit to descartes, inspired by the moveitstateadapter, i.e.:
            `std::string group_name = req.group_name;
        robot_state::RobotState start_state;
        robot_state::robotStateMsgToRobotState(req.start_state, start_state);
        moveit::core::RobotModelConstPtr robot_model = robot_state->getRobotModel();
        const moveit::core::JointModelGroup* joint_model_group_ptr = robot_state->getJointModelGroup(group_name);
        const std::vector<std::string>& link_names = joint_model_group_ptr->getLinkModelNames();
        tool_frame = link_names.back();
        world_frame = robot_state->getRobotModel()->getModelFrame();`

For the GetCartesianPath's req.waypoints, I've translated them into a vector of shared_ptr to descartes_core::TrajectoryPt, inspired from godel/pathplannig. 

Further, How can the remaining request part of teh GetCartesianPath srv be utilized?  
 `float64 max_step   
 float64 jump_threshold   
 bool avoid_collisions           
 moveit_msgs/Constraints path_constraints`

jrgnicho:
@madratman  Thank you for your contribution, I'll take a look at it soon.

madratman:
Thanks. Although, I am afraid it isn't useful as of now. One more thing - talking about interpolators, there are many possible ways  -  linear, parabolic and cubic blends, bezier curves, right? Are there any plans to implement them? 
Existing/deprecated repos:
[industrial_trajectory_filters](http://wiki.ros.org/action/fullsearch/industrial_trajectory_filters?action=fullsearch&context=180&value=linkto%3A%22industrial_trajectory_filters%22)
[...]

Issue #51, opened 2015-02-06
Interpolator Class


jrgnicho:
The current descartes planner (PlanningGraph, SparsePlanner) only operate on dense trajectories.  Thus an Interpolator would produce intermediate points between the sparse waypoints and then pass the interpolated dense trajectory to a planner.  In addition, it should allow for several interpolation strategies (cartesian, joint, etc) and tolerances (waypoint zone) for each waypoint.

shaun-edwards:
@jrgnicho, we need more detail on this one.  I'd like to point a few people to this bug, but there needs to be more details.  What's the input?  What's the output?  What data structures are envisioned?  Can we point to existing robot languages as an example?

Jmeyer1292:
While useful, I feel that this sort of functionality belongs to a higher level package that wraps descartes as it is now.

PR #65, opened 2015-03-07
Velocity pruning

Jmeyer1292:
Related to issue #63 and Issue #48, here is my first crack at adding time windowing data to our trajectory points. First off, this pull-request is not quite ready because there are many documentation changes that need to be made but I would like to get eyes on this asap.

This PR replaces the Point Transition class with a timing constraint struct that has an upper and lower limit for time. Only the upper limit on time is currently used or checked (I'm not sure if the lower is needed). This timing must be specified when building new types of Trajectory Points but it can be default constructed which puts the upper and lower limits to zero.

Planning graph has been changed to incorporate pull request #64 as well as to remove the absolute joint delta check in place of an updated function in RobotModel that takes two joint positions and a time delta and sees if motion between those two points is possible in that window of time.

Again this is under development, but I want to see if people think I'm on the track.


ClayFlannigan:
Something for us to consider: a fully specified joint state would specify the position and velocity for each point.  Acceptable trajectories would require less than than maximum joint velocities and accelerations between points.  The current planner only considers positions and ignores dynamics (velocities and accelerations).  This pull adds time constraints thereby permitting calculation of average velocities along a trajectory.  In the limit as the step size approaches zero, the average and the instantaneous velocities are the same.  The question is: should we instead specify velocities at points, thereby permitting us to calculate accelerations to achieve the trajectory?  Or should we look at the second (numerical) derivative to calculate accelerations?  

UPDATE: I think the above is flawed, because we can't know the joint velocities ahead of time.  All we know is time, so calculating the average velocity is the best we can do.  I guess there is still the question of whether we should also calculate the average acceleration?

gavanderhoorn:
Perhaps also something to keep in mind (or to purposefully ignore): a lot (some?) industrial robot controllers work with trajectory segments, where you can specify a target velocity to be reached over that segment, not at the end point. With blending/falling through/contouring/x, you can get the controller to smoothly stitch segments together, but at the cost of accuracy (in both time and space, I believe). Making the segments small enough can work around this in some way, but (in most cases) at the cost of maximum attainable performance.

I've always considered this more position control with velocity constraints over traj segments than actually being able to define the state of the manipulator (pose & velocity) at specific points. ROS generated trajectories seem to assume we can actually do the latter. I think only the `motoman` driver is currently capable of that within ros-i.

Is this something we want to keep in mind? Or is `descartes` being designed _for the future_, where all industrial controllers provide full access to their motion control systems?

ClayFlannigan:
I agree with @gavanderhoorn in that it is probably best to consider it as position control with velocity constraints and rather than a true velocity controller, so the current approach seems like the pragmatic and right one.  

Related, a filter could be applied post-planning that generates a smooth velocity trajectory.  Here is the classic cubic spline equation which requires as inputs pos_0, pos_1, t_0, t_1, and vel_0 to calculate the smooth velocity at the trajectory endpoint, vel_1. A similar approach could be used if we wanted to specify joint velocities.

from: Spong et. al., "Robot Modeling and Control", 2005
cubic_interpolated_trajectory

descartes_core/include/descartes_core/trajectory_timing_constraint.h
```diff
@@ -0,0 +1,52 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/*
+ * trajectory_pt_transition.h
+ *
+ *  Created on: Jun 5, 2014
+ *      Author: Dan Solomon
```
shaun-edwards: Update author.

descartes_moveit/include/descartes_moveit/moveit_state_adapter.h
```diff
@@ -48,7 +47,8 @@ class MoveitStateAdapter : public descartes_core::RobotModel
    * @param world_frame work object frame name
    */
   MoveitStateAdapter(const moveit::core::RobotState & robot_state, const std::string & group_name,
-                    const std::string & tool_frame, const std::string & world_frame,
+                     const std::string & tool_frame, const std::string & world_frame,
+                     const std::vector<double>& velocity_limits,
```
shaun-edwards: Aren't the velocity limits stored in the `moveit::core::RobotState`?

descartes_core/include/descartes_core/robot_model.h
```diff
@@ -102,6 +102,17 @@ class RobotModel
...
 
+  /**
+   * @brief Performs necessary checks to see if the robot is capable of moving from the initial joint pose
+   *        to the final pose in dt seconds
+   * @param  from_joint_pose [description]
+   * @param  to_joint_pose   [description]
+   * @param  dt              [description]
+   * @return                 [description]
+   */
+  virtual bool isValidMove(const std::vector<double>& from_joint_pose, const std::vector<double>& to_joint_pose,
```
shaun-edwards, line 131: Should this method name be more specific?  It really only checks velocity limits?  The name implies more.

descartes_moveit/src/moveit_state_adapter.cpp
```diff
@@ -33,13 +33,15 @@ namespace descartes_moveit
 
 MoveitStateAdapter::MoveitStateAdapter()
 {
-
+  velocity_limits_ = std::vector<double>(7, 0.1);
```
shaun-edwards: This should be removed if the MoveIt robot state already contains this info.

descartes_planner/src/planning_graph.cpp
```diff
@@ -41,7 +41,7 @@ using namespace descartes_trajectory;
 namespace descartes_planner
 {
 
-const double MAX_JOINT_DIFF = M_PI;
+const double MAX_JOINT_DIFF = M_PI / 20;
```
shaun-edwards: This can go away now...right?

descartes_planner/src/planning_graph.cpp
```diff
@@ -595,39 +595,81 @@ bool PlanningGraph::removeTrajectory(TrajectoryPtPtr point)
...
-      start_points.push_back(jv);
+      cart_id = c_iter->first;
+      break;
     }
   }
+
+  // Test for error case
+  if (cart_id.is_nil())
+  {
+    ROS_ERROR("Could not locate TrajectoryPt with nil previous point. Graph may be cyclic.");
```
shaun-edwards, line 586: Should this ever happen?  Maybe a comment to describe when it does (if it could)?

descartes_planner/src/planning_graph.cpp
```diff
@@ -595,39 +595,81 @@ bool PlanningGraph::removeTrajectory(TrajectoryPtPtr point)
...
-      end_points.push_back(jv);
+      cart_id = c_iter->first;
+      break;
     }
   }
+
+  // Test for error case
+  if (cart_id.is_nil())
+  {
+    ROS_ERROR("Could not locate TrajectoryPt with nil next point. Graph may be cyclic.");
```
shaun-edwards, line 626: See comment above.

descartes_planner/src/planning_graph.cpp
```diff
@@ -1004,19 +1058,17 @@ double PlanningGraph::linearWeight(JointTrajectoryPt start, JointTrajectoryPt en
     std::vector<double> end_vector = joint_poses_end[0];
     if (start_vector.size() == end_vector.size())
     {
+      if (end.getTiming().isSpecified() && !robot_model_->isValidMove(start_vector, end_vector, end.getTiming().upper_))
+      {
+        return MAX_EXCEEDED_PENALTY;
```
shaun-edwards: Am I reading this correctly...if the timing is specified and the move is not valid, we set the max penalty?  Shouldn't we prune the edge?  Or is this a "magic" value that indicates we should prune the edge?

descartes_core/include/descartes_core/trajectory_timing_constraint.h
```diff
@@ -0,0 +1,52 @@
...
+
+#ifndef TRAJECTORY_TIMING_CONSTRAINT_H
+#define TRAJECTORY_TIMING_CONSTRAINT_H
+
+namespace descartes_core
+{
+  /**
+   * Timing constraints 
+   */
+  struct TimingConstraint
```
shaun-edwards, line 42: Added constructor for a timing constraint with a single value (i.e. just a nominal timing).

shaun-edwards:
@gavanderhoorn, @ClayFlannigan, it sounds like you are OK with the current approach.  I would like provide a little bit of the total vision (this will eventually make it's way to the REP).  When considering trajectories, Descartes will contain two types of trajectories:
- Dense trajectory (currently implemented) - A dense trajectory is defined point by point.  A trajectory point (joint/cart/whatever) will contain a pose, with associated velocity, acceleration, and timing.  I realize this could be over/under constrained depending how many derivatives you think are required.  A key requirement is that motion in between trajectory points is well defined (i.e. smooth and collision free).
- Sparse trajectory (in progress, see start in PR #51) - This could be any number of representations.  It could look like traditional robot code (as @gavanderhoorn pointed out), gcode, or anything.  In general, these trajectories will contain points and segment constrains (linear, circular, joint, velocity, timing, etc...).  These representations are efficient, but there is no guarantee that motion segments are well behaved and/or is collision free.  Sparse trajectories must be converted to dense trajectories for planning.  This conversion depends on the representation used, but should result in populated velocity, acceleration, and timing values.

Initially @Jmeyer1292 and I chose timing because it generalizes to any kind of point.  Therefore, it can exist in the base `TrajectoryPt` type.  I believe this was Clay's conclusion as well.  The one possible downside to this change implies a direction to the motion (i.e. a `delta t` from the previous point).  A `delta t` approach was chosen to allow an upper and lower bound to be specified (kind of like the pose constraints).  Not sure how to avoid the implied direction of motion.

To @ClayFlannigan 's second point, the joint trajectory that results from planning will have to be smoothed.  If the graph used for planning does not contain edges or nodes that are physically impossible, then a smooth trajectory through the planned trajectory is likely possible.  Unfortunately, there are some trajectory conditions, like path reversals, which are not considered at pruning time.  This is a known limitation at this time.

One final thought, @ClayFlannigan and I discussed using instantaneous velocities at a point to determine if the point (and therefore the node) is valid.  Using this PR's method of pruning edges is good, but removing a node entirely is much better.  I believe I understand how to generalize this to both cart/joint points.  This will be a future PR.

descartes_core/include/descartes_core/robot_model.h
Jmeyer1292, line 131: I was originally thinking that the method might be a general test for the validity of a motion and the specifics would be up to the implementation. That way we could incorporate acceleration or collisions, for example.

Admittedly that's probably not going to happen, so would something like checkVelocityConstraints() more apt?

descartes_planner/src/planning_graph.cpp
Jmeyer1292, line 586: The addTrajectory function of planning_graph takes a pointer and a previous and next id. One could add arbitrary edges with this.

madratman:
For smoothing, can [this](http://wiki.ros.org/trajectory_filters) be useful?
[More details in fourth comment in issue #52](https://github.com/ros-industrial-consortium/descartes/issues/52)

shaun-edwards:
@Jmeyer1292, this PR can't be merged automatically...can you rebase to fix this?

Jmeyer1292:
@shaun-edwards I rebased the code, but there is still the outstanding issue of making the timing work with the sparse trajectory planner. I'll look into that shortly.

descartes_core/include/descartes_core/robot_model.h
Jmeyer1292, line 131: ping

Jmeyer1292:
@shaun-edwards or @jrgnicho If y'all would take a look at this again when you have a chance, I'd appreciate it.

shaun-edwards:
@jrgnicho, please review

descartes_core/include/descartes_core/trajectory_timing_constraint.h
```diff
@@ -0,0 +1,80 @@
...
+      , upper_(upper)
+    {}
+
+    /**
+     * @brief Checks if the given timing constraint has been specified
+     * @return true if the point is specified
+     */
+    bool isSpecified() const { return upper_ > 0.0; }
+
+    double lower_, upper_;
```
jrgnicho: Struct naming convention [see here](http://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Variable_Names)

descartes_moveit/src/moveit_state_adapter.cpp
```diff
@@ -315,5 +357,29 @@ int MoveitStateAdapter::getDOF() const
...
+  for (std::vector<double>::size_type i = 0; i < from_joint_pose.size(); ++i)
+  {
+    if ( std::abs(from_joint_pose[i] - to_joint_pose[i]) > max_joint_deltas[i] )
+    {
+      return false;
+    }
+  }
+
+  return true;
+}
```
jrgnicho, line 397: Maybe we should add a check for same size arrays

descartes_planner/src/dense_planner.cpp
```diff
@@ -97,6 +96,7 @@ bool DensePlanner::updatePath()
       || (std::find_if(cart_map.begin(),cart_map.end(),predicate) == cart_map.end())
       || first_id == descartes_core::TrajectoryID::make_nil())
   {
+    ROS_ERROR_STREAM("Failure case 1");
```
jrgnicho: What is a "Failure case 1" error?

descartes_planner/src/dense_planner.cpp
```diff
@@ -137,11 +138,13 @@ bool DensePlanner::updatePath()
...
     {
+      ROS_ERROR_STREAM("Failure case 2");
       error_code_ = descartes_core::PlannerError::INCOMPLETE_PATH;
     }
   }
   else
   {
+    ROS_ERROR_STREAM("Failure case 3");
     error_code_ = descartes_core::PlannerError::IK_NOT_AVAILABLE;
   }
```
jrgnicho, line 146: Can the error message be made more specific to the type of failure?

descartes_planner/src/planning_graph.cpp
```diff
@@ -672,7 +675,7 @@ bool PlanningGraph::getShortestPath(double &cost, std::list<JointTrajectoryPt> &
 
     for (std::list<JointGraph::vertex_descriptor>::iterator end = end_points.begin(); end != end_points.end(); end++)
     {
-      //ROS_INFO("Checking path: S[%d] -> E[%d]", *start, *end);
+      // ROS_INFO("Checking path: S[%d] -> E[%d]", *start, *end);
```
jrgnicho: This should be turned into a ROS_DEBUG .. or be removed

descartes_planner/src/sparse_planner.cpp
```diff
@@ -730,6 +758,22 @@ int SparsePlanner::interpolateSparseTrajectory(const SolutionArray& sparse_solut
...
+
+          // look up previous points joint solution
+          const JointTrajectoryPt& last_joint_pt = joint_points_map_.at(cart_points_[pos-1]->getID());
+          std::vector<double> sol;
+          last_joint_pt.getNominalJointPose(std::vector<double>(), *robot_model, sol);
+          // look up dt
+          double dt = timing_cache_[pos].upper_;
+          // check validity of joint motion
+
+          if (!robot_model->isValidMove(sol, aprox_interp, dt))
```
jrgnicho: I think this **isValidMove** check makes the **checkJointChanges** call above unnecessary.

descartes_moveit/src/moveit_state_adapter.cpp
Jmeyer1292, line 397: What's our policy on using asserts?

descartes_planner/src/sparse_planner.cpp
Jmeyer1292: Well, it seems to me that `checkJointChanges` tests the interpolated joint position against the valid IK for that point. `isValidMove` checks to see if that point is achievable from the previous one given the joint velocity constraints of the system.

descartes_core/include/descartes_core/trajectory_timing_constraint.h
```diff
@@ -0,0 +1,80 @@
...
+    TimingConstraint(double lower, double upper)
+      : lower(lower)
+      , upper(upper)
+    {}
+
+    /**
+     * @brief Checks if the given timing constraint has been specified
+     * @return true if the point is specified
+     */
+    bool isSpecified() const { return upper > 0.0; }
```
jrgnicho, line 94: Should we check for negative values on the **lower** member as well?

descartes_moveit/src/moveit_state_adapter.cpp
```diff
@@ -26,6 +26,37 @@
...
+{
+  std::vector<double> result;
+
+  auto models = state.getJointModelGroup(group_name)->getActiveJointModels();
+  for (const moveit::core::JointModel* model : models)
+  {
+    const auto& bounds = model->getVariableBounds();
+    // Check to see if there is a single bounds constraint (more might indicate
+    // not revolute joint)
+    if (bounds.size() != 1)
```
jrgnicho: I think it be more reliable to check the [type](http://docs.ros.org/indigo/api/moveit_core/html/classmoveit_1_1core_1_1JointModel.html#a17a4c66425888df27c53c84bb1fa0585) to determine if they are either revolute or prismatic joints.

descartes_moveit/src/moveit_state_adapter.cpp
```diff
@@ -320,5 +363,36 @@ int MoveitStateAdapter::getDOF() const
...
+                                     const std::vector<double>& to_joint_pose,
+                                     double dt) const
+{
+  std::vector<double> max_joint_deltas;
+  max_joint_deltas.reserve(velocity_limits_.size());
+
+  // Check for equal sized arrays
+  if (from_joint_pose.size() != to_joint_pose.size())
+  {
+    ROS_DEBUG_STREAM("To and From joint poses are of different sizes.");
```
jrgnicho: This should probably be a ROS_ERROR

descartes_moveit/src/moveit_state_adapter.cpp
jrgnicho, line 397: We don't really have a policy for this but I suppose that it is justifiable if the application runs into an irrecoverable condition.  My preference is printing a ROS_ERROR and returning false.

descartes_planner/src/sparse_planner.cpp
```diff
@@ -23,11 +23,28 @@
...
+
+      for (const auto& ptr : points)
+      {
+        result.push_back(ptr->getTiming());
+      }
+      return result;
+    }
+}
+
+
```
jrgnicho, line 47: Would it not be a bit more efficient to allow for an output argument for the vector instead of returning a copy?

descartes_planner/src/sparse_planner.cpp
```diff
@@ -131,6 +148,15 @@ bool SparsePlanner::planPath(const std::vector<TrajectoryPtPtr>& traj)
   }
 
   ros::Time start_time = ros::Time::now();
+
+  // Save the timing information
+  timing_cache_ = extractTiming(traj);
+  // Unconstrian the times for the sparse sampling
```
jrgnicho: typo Unconstrian

jrgnicho:
Code reviews aren't fun.  Just a few minor comments but in general this one is ready to merge

descartes_moveit/src/moveit_state_adapter.cpp
Jmeyer1292: +1

descartes_planner/src/sparse_planner.cpp
Jmeyer1292, line 47: I hesitate to make general comments on compiler performance without benchmarking, but I doubt that output arguments are any faster because of return-value-optimizations and the fact that in c++-11 values returned from functions are implicitly moved.

descartes_planner/src/sparse_planner.cpp
Jmeyer1292: +1

descartes_moveit/src/moveit_state_adapter.cpp
Jmeyer1292: +1 I did that on the IK seed generator, but wasn't aware of it when I started with this

descartes_core/include/descartes_core/trajectory_timing_constraint.h
Jmeyer1292, line 94: I'm more inclined to put such a check in as a ROS_ASSERT into the constructor. I did not mean this function as a validity check but a test for whether or not the user actually specified a dt.

Jmeyer1292:
@jrgnicho I removed the assertion and replaced it with a warning. In the event of invalid (read negative) timing values, the times are instead set to zero indicating unspecified.

jrgnicho:
+1

Issue #67, opened 2015-03-09
core: build script hardcodes cmake build type (set to debug)


gavanderhoorn:
As per title. See [descartes_core/CMakeLists.txt](https://github.com/ros-industrial-consortium/descartes/blob/672b9c0907755ef23bfc92905942425cc31e19e0/descartes_core/CMakeLists.txt#L4).

Not sure if this is _by design_, but it is considered bad practice to set the CMake build type from within a `CMakeLists.txt` file. Recommended is to do it at configuration time (ie: `catkin_make -DCMAKE_BUILD_TYPE=...`).

shaun-edwards:
@jrgnicho, this is a high priority fix.  We should also see if this affects `amos`.  @Jmeyer1292, this might have an affect on `godel`.

gavanderhoorn:
Fixed in #68.

PR #68, opened 2015-03-09
Removed debug build flag

jrgnicho:
This is the fix to issue #67.
@shaun-edwards, please review


shaun-edwards:
+1..Thanks for the quick turn-around...thanks @gavanderhoorn for finding this.

Issue #66, opened 2015-03-09
all: pkgs using C++11 features should export their cflags (and check compiler support)


gavanderhoorn:
All pkgs in the `descartes` repository add the `-std=gnu++0x` compiler option somewhere in their `CMakeLists.txt` ([descartes_core](https://github.com/ros-industrial-consortium/descartes/blob/672b9c0907755ef23bfc92905942425cc31e19e0/descartes_core/CMakeLists.txt#L5) fi). Dependant pkgs also need to set this flag, in order to be able to use headers from `descartes` ([godel_path_planning](https://github.com/ros-industrial-consortium/godel/blob/eea4ad584052ed29442d6b29bea1a8c0c27430e7/godel_path_planning/CMakeLists.txt#L20) fi).

It is customary in such cases (in `catkin` projects) for the providing pkg to export any necessary `cflags` by exporting a `pkgname_CFLAGS` or `pkgname_DEFINITIONS` variable (with a preference for the latter, it seems) in a `.cmake` snippet that is passed to dependants using the `CFG_EXTRAS` argument to the `catkin_package(..)` call.

All pkgs in `descartes` should use this mechanism to export the required `-std=gnu++0x`.

---

Related: afaict, the `-std=gnu++0x` option could actually be GCC specific and should ideally be guarded by a check for `CMAKE_COMPILER_IS_GNUCXX` (although apparently CLang is compatible with GCC command line options, so it should work there).

gavanderhoorn:
Alternatively, the `.cmake` snippet could append the option directly to the `CMAKE_CXX_FLAGS` variable itself, which would avoid the situation where dependants forget to add an `add_definitions(${descartes_core_DEFINITIONS})` line to their `CMakeLists.txt`.

I don't like that too much though, as simply `find_package(..)`-ing the pkgs would then lead to the flags being updated automatically, which is probably not what the user wanted. Additionally, there are several buildtype-specific variants of `CMAKE_CXX_FLAGS`, which would require some if-else work to figure out which one to update from the snippet.

jrgnicho:
I'm not too familiar with the use of .cmake snippets for exporting package specific flags, could these flags (- std=gnn++0x in this case)  not be defined and exported from the CMakeList.txt file in the package?

gavanderhoorn:
I'll submit a PR later.

Re: setting them from the `CMakeLists.txt`: no, since that file is not imported by the dependants. The `CFG_EXTRAS` snippet is automatically imported.

PR #69, opened 2015-03-10
Fix for issue #66: force dependant pkgs to enable C++0x/11 support using CFG_EXTRAS

gavanderhoorn:
As per subject.

Dependants `find_package(..)`-ing any package in `descartes` will see the following message printed in the CMake configuration phase:

```
-- descartes_core: enabling C++0x (required)
```

Contrary to what I wrote in #66, I opted to directly append the flag to the `CMAKE_CXX_FLAGS`, as it is always used irrespective of build type, and avoids users forgetting to add the exported definitions in their `CMakeLists.txt` (although that would lead to highly visible compiler errors).

I also did not export the `-std=c++0x` flag in every `descartes` pkg, as it is impossible to depend on any of `descartes`' pkgs without also depending on `descartes_core`, which already exports that flag.

Finally, I think it's considered bad practice to rely on a dependency for setting option flags in dependants, so having this exported does _not_ mean that dependants should not enable C++0x/11 support by themselves if they need it for their own sources (so pkgs in `godel` should still enable it in their `CMakeLists.txt` if they need it).

This is only meant to make sure that dependants do not have to worry about what flags `descartes` needs, if they use it.

### DIFF
--- a/descartes_core/CMakeLists.txt
+++ b/descartes_core/CMakeLists.txt
@@ -1,7 +1,6 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(descartes_core)
 
-add_definitions(-std=gnu++0x)
 find_package(catkin REQUIRED COMPONENTS
   console_bridge
   moveit_core
@@ -22,6 +21,8 @@ catkin_package(
     Boost
     console_bridge
     Eigen
+  CFG_EXTRAS
+    descartes_core-extras.cmake
 )
 
 ###########
--- a/descartes_core/cmake/descartes_core-extras.cmake
+++ b/descartes_core/cmake/descartes_core-extras.cmake
@@ -1,0 +1,15 @@
+#
+# descartes_core CMake/catkin extras snippet
+#
+# Enable (GCC) C++0x/11 support by adding the corresponding compiler flag
+# directly.
+#
+# TODO: for newer versions of CMake, this should probably be done using
+#       add_compile_options(..) (>= 2.8.11), or target_compile_features(..)
+#       (>= 3.1.x).
+#
+if (NOT descartes_core_FIND_QUIETLY)
+  message(STATUS "descartes_core: enabling C++0x (required)")
+endif()
+
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++0x")
--- a/descartes_moveit/CMakeLists.txt
+++ b/descartes_moveit/CMakeLists.txt
@@ -1,7 +1,6 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(descartes_moveit)
 
-add_definitions(-std=gnu++0x)
 find_package(catkin REQUIRED COMPONENTS
   moveit_core
   tf
@@ -15,7 +14,6 @@ find_package(catkin REQUIRED COMPONENTS
 find_package(rosconsole_bridge REQUIRED)
 find_package(Boost REQUIRED)
 find_package(Eigen REQUIRED)
-
 
 catkin_package(
   INCLUDE_DIRS
--- a/descartes_planner/CMakeLists.txt
+++ b/descartes_planner/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(descartes_planner)
-add_definitions(-std=gnu++0x)
+
 find_package(catkin REQUIRED COMPONENTS
   descartes_core
   descartes_trajectory
@@ -9,6 +9,7 @@ find_package(catkin REQUIRED COMPONENTS
   roscpp
   pluginlib
 )
+
 find_package(Boost REQUIRED)
 find_package(Eigen REQUIRED)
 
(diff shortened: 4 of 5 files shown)


jrgnicho:
Will exporting this flag would mean that it automatically applies it to every package that has build dependencies on descartes?

gavanderhoorn:
Yes, that is the idea. All pkgs that depend on any pkg in `descartes` will have `-std=c++0x` appended to their `CXX_FLAGS`.

btw: if descartes does not expose any C++0x features in any of its headers, then this might actually be unnecessary.

jrgnicho:
While some build errors that result from lack of c++0x friendly code are very obvious, there are others that might be difficult to identify.  One such case is that of namespace ambiguities that prevent the compiler from resolving library artifacts  named equally under different namespaces.  The error messages for this failure are mostly unreadable and provide minimal cue for the real cause of the error.  When this happened to me I was able to identify the problem because I had just manually added the **-std=gnu++0x** definition.; however the changes in this PR hide the step that leads to this error.  I'm not sure that there be a good way of alerting of the potential consequences of linking against descarte

gavanderhoorn:
I'm a bit confused: are you hesitant to merge this because it may complicate debugging compiler errors, or ? It would seem that any 'hidden' additions to the compilation flags would complicate that, but I do see where enabling C++0x is somewhat of a special case.

Note btw that currently this only appends the `-std=c++0x` flag, but the same setup could be used to export other build flags. Relying on the user to set such flags in his own `CMakeLists.txt` is not very nice, they will forget and every pkg using `descartes` will have to repeat them (same as with the `-DROS=1` and `-DLINUXSOCKETS=1` in `simple_message` & `industrial_robot_client`).

jrgnicho:
Yes, that is my worry.  Perhaps printing status messages  during the build will be enough to help address such build errors if they were to happen.
+1 for merge

Jmeyer1292:
Could we not just go through and remove the use of c++11 in the header files of Descartes? I think we're only using enum classes, although I need to look again.

Or might this cause some kind of ABI issue if this is compiled with c++11 and other pieces are not?

gavanderhoorn:
@Jmeyer1292: yes, that is also an option.

But that would just postpone the issue, imo.

shaun-edwards:
Back in my day...we landed people on the moon with just plain old C.  We didn't even need the `++`.  Aside from my distrust of any change...is there a reason not to use C++ 11?  I believe even ROS is transitioning to C++ 11.  Just wondering?

gavanderhoorn:
off-topic: [Apollo Guidance Computer (AGC) Simulator](http://svtsim.com/moonjs/agc.html). Personally I prefer the way we do things now ;)

jrgnicho:
The only problem that I have encountered is that of the namespace collisions that might occur when enabling c++0x on a package that wasn't developed with the new language features.  However, I think this is a rare occurrence.

Jmeyer1292:
12.04 uses gcc 4.6 which does not have a fully-conforming c++11 implementation. We have to be conscious of that when developing or we will break code for people on Hydro.

https://gcc.gnu.org/projects/cxx0x.html

gavanderhoorn:
@Jmeyer1292: yes, that is what I was concerned about when I asked about this earlier (when discussing godel iirc).

@jrgnicho: which collisions are you refering to exactly?

jrgnicho:
There was a **isnan** method defined in the ikfast.h header which is now defined under the **std** namespace in c++11.  This is the ambiguity that the compiler could not resolve.

gavanderhoorn:
Is that `ikfast.h` used by `moveit_ikfast`? Or some other file? Cause I can't find it in [moveit_ikfast/templates/ikfast.h](https://github.com/ros-planning/moveit_ikfast/blob/370a733118086109ebe86f655aff3ace93ab7614/templates/ikfast.h).

If it is that file, could you log an issue? It would probably be a good idea to either call `::isnan(..)` or `std::isnan(..)` then.

jrgnicho:
It is actually in one the files that get generated when generating a moveit kinematics plugin [see here](https://github.com/ros-industrial-consortium/godel/blob/hydro-devel/motoman_sia20d_ikfast_manipulator/include/motoman_sia20d_ikfast_manipulator/motoman_sia20d_manipulator_ikfast_solver.hpp#L72).  Normally one wouldn't link against it but we did it because there was some functionality that we wanted to exposed that isn't accessible through the plugin.  This is such a rare case that logging an issue for it might be unnecessary.

gavanderhoorn:
Ah, that's not `ikfast.h` :).

Perhaps log the issue against [rdiankov/openrave](https://github.com/rdiankov/openrave)? That is where the file is generated (in [openrave/python/ikfast_generator_cpp.py](https://github.com/rdiankov/openrave/blob/2f7c8a6e582fd74746902e63b5707052fee87b09/python/ikfast_generator_cpp.py)).

jrgnicho:
+1 for merge

gavanderhoorn:
Some more info on [C++0x/C++11 Support in GCC](https://gcc.gnu.org/projects/cxx0x.html): [GCC-4.6](https://gcc.gnu.org/gcc-4.6/cxx0x_status.html) vs [GCC-4.8](https://gcc.gnu.org/gcc-4.8/cxx0x_status.html).

Issue #70, opened 2015-03-16
Inconsistent Arguments to RobotModel initialize method


Jmeyer1292:
This is a minor quibble, but the current RobotModel initialize method takes the robot description string by value instead of by reference. This appears to be a simple typo, but I wanted to check.

https://github.com/ros-industrial-consortium/descartes/blob/hydro-devel/descartes_core/include/descartes_core/robot_model.h#L102

jrgnicho:
@Jmeyer1292  can this be closed?

PR #71, opened 2015-03-16
Robot Description Argument Taken By Reference (Issue #70)

Jmeyer1292:
This is a small fix going inline with issue #70. It changes the robot_description argument to by taken by reference like the other arguments of the Initialize function.


Jmeyer1292:
Done

shaun-edwards:
+1.

PR #72, opened 2015-03-17
Fix indigo build issues

jrgnicho:
@shaun-edwards, the changes in this PR fix the build errors in ros-indigo.  You could either merge this and/or use it as a departure point for the **indigo-devel** branch.

### DIFF
--- a/descartes_core/CMakeLists.txt
+++ b/descartes_core/CMakeLists.txt
@@ -3,14 +3,13 @@ project(descartes_core)
 
 add_definitions(-std=gnu++0x)
 find_package(catkin REQUIRED COMPONENTS
-  console_bridge
   moveit_core
   roscpp
+  cmake_modules
 )
 
 find_package(Boost REQUIRED)
 find_package(Eigen REQUIRED)
-
 
 catkin_package(
   INCLUDE_DIRS
@@ -20,7 +19,6 @@ catkin_package(
     moveit_core
   DEPENDS
     Boost
-    console_bridge
     Eigen
 )
 
--- a/descartes_core/package.xml
+++ b/descartes_core/package.xml
@@ -17,6 +17,7 @@
   <build_depend>moveit_core</build_depend>
   <build_depend>roscpp</build_depend>
   <build_depend>cmake_modules</build_depend>
+  <build_depend>eigen</build_depend>
 
   <run_depend>boost</run_depend>
   <run_depend>moveit_core</run_depend>
--- a/descartes_core/package.xml
+++ b/descartes_core/package.xml
@@ -14,12 +14,11 @@
   <buildtool_depend>catkin</buildtool_depend>
 
   <build_depend>boost</build_depend>
-  <build_depend>console_bridge</build_depend>
   <build_depend>moveit_core</build_depend>
   <build_depend>roscpp</build_depend>
+  <build_depend>cmake_modules</build_depend>
 
   <run_depend>boost</run_depend>
-  <run_depend>console_bridge</run_depend>
   <run_depend>moveit_core</run_depend>
   <run_depend>roscpp</run_depend>
 
--- a/descartes_planner/CMakeLists.txt
+++ b/descartes_planner/CMakeLists.txt
@@ -4,10 +4,10 @@ add_definitions(-std=gnu++0x)
 find_package(catkin REQUIRED COMPONENTS
   descartes_core
   descartes_trajectory
-  console_bridge
   moveit_core
   roscpp
   pluginlib
+  cmake_modules
 )
 find_package(Boost REQUIRED)
 find_package(Eigen REQUIRED)
@@ -24,9 +24,10 @@ catkin_package(
   CATKIN_DEPENDS
     roscpp
     moveit_core
+    descartes_core
+    descartes_trajectory
   DEPENDS
     Boost
-    console_bridge
     Eigen
 )
 
--- a/descartes_trajectory/CMakeLists.txt
+++ b/descartes_trajectory/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 2.8.3)
 project(descartes_trajectory)
 add_definitions(-std=gnu++0x)
 find_package(catkin REQUIRED COMPONENTS
-  console_bridge
+  cmake_modules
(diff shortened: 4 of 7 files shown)


gavanderhoorn:
fyi: `cmake_modules` is fully (bw-)compatible with Hydro.

descartes_core/package.xml
```diff
@@ -17,11 +17,13 @@
   <build_depend>console_bridge</build_depend>
   <build_depend>moveit_core</build_depend>
   <build_depend>roscpp</build_depend>
+  <build_depend>cmake_modules</build_depend>
 
   <run_depend>boost</run_depend>
   <run_depend>console_bridge</run_depend>
   <run_depend>moveit_core</run_depend>
   <run_depend>roscpp</run_depend>
+  <run_depend>cmake_modules</run_depend>
```
gavanderhoorn: I don't believe a `run_depend` is needed with `cmake_modules`. It's a build-only thing.

jrgnicho: Ok, I'll list cmake_modules as a build_depend only

jrgnicho:
The latest commits mainly removed the **console_bridge** dependencies.  This branch builds in both hydro and indigo.  @gavanderhoorn I tried removing the **cmake_modules** run dependency but it game me the following build error:

```
CMake Error at /opt/ros/indigo/share/catkin/cmake/catkin_package.cmake:217 (message):
  catkin_package() DEPENDS on the catkin package 'cmake_modules' which must
  therefore be listed as a run dependency in the package.xml
```

gavanderhoorn:
@jrgnicho: yes, but `cmake_modules` should not be added to `CATKIN_DEPENDS` either, afaik. See [catkin_package() - CATKIN_DEPENDS](http://docs.ros.org/hydro/api/catkin/html/dev_guide/generated_cmake_api.html#catkin-package) on the _Extracted CMake API reference_ page:

> [..] Therefore their INCLUDE_DIRS and LIBRARIES will be appended to ours. [..]

`cmake_modules` doesn't have any headers or libraries.

And you can also just `find_package(..)` it on its own, no need to add it to the `find_package(catkin ..)` line.

jrgnicho:
@gavanderhoorn The last commits removes the cmake_modules run dependencies.  It still builds on both hydro and indigo

descartes_core/CMakeLists.txt
```diff
@@ -3,13 +3,12 @@ project(descartes_core)
...
 add_definitions(-std=gnu++0x)
 find_package(catkin REQUIRED COMPONENTS
-  console_bridge
   moveit_core
   roscpp
+  cmake_modules
 )
 
 find_package(Boost REQUIRED)
-find_package(Eigen REQUIRED)
```
gavanderhoorn, line 12: Why is this removed if `Eigen` is listed as a `DEPENDS` in `catkin_package(..)` further on (and `Eigen_INCLUDE_DIRS` is also referenced)?

Also: if you don't use `Eigen` in this package, then it would seem there is also no need for `cmake_modules`?

descartes_planner/CMakeLists.txt
```diff
@@ -24,9 +22,10 @@ catkin_package(
   CATKIN_DEPENDS
     roscpp
     moveit_core
+    descartes_core
+    descartes_trajectory
   DEPENDS
     Boost
-    console_bridge
     Eigen
```
gavanderhoorn: See earlier comment.

descartes_trajectory/CMakeLists.txt
```diff
@@ -2,14 +2,12 @@ cmake_minimum_required(VERSION 2.8.3)
...
 add_definitions(-std=gnu++0x)
 find_package(catkin REQUIRED COMPONENTS
-  console_bridge
   descartes_core
   moveit_core
   roscpp
 )
 
 find_package(Boost REQUIRED)
-find_package(Eigen REQUIRED)
```
gavanderhoorn, line 12: See earlier comment.

gavanderhoorn:
If I understand things correctly, the `boost` and `eigen` keys should be added as `build_depend`s, and a `run_depend` on `boost` is necessary (but only if any of its libraries are used).

For `Eigen` there are actually three `rosdep` keys: [eigen](http://rosindex.github.io/d/eigen/), [eigen2](http://rosindex.github.io/d/eigen2/) and [eigen3](http://rosindex.github.io/d/eigen3/). For `Boost` there is only one: [boost](http://rosindex.github.io/d/boost/).

descartes_core/CMakeLists.txt
jrgnicho, line 12: The Eigen library is used in this package.  Wouldn't **find(catkin REQUIRED COMPONENTS cmake_modudes ...)** include Eigen as well;  in which case calling find_package(Eigen ...) would be redundant.  If this is so then adding Eigen_INCLUDE_DIRS or Eigen under DEPENDS isn't required either.

gavanderhoorn, line 12: Nope. You have to `find_package(Eigen ..)` yourself. `cmake_modules` just provides the find script.

jrgnicho, line 12: I'm surprised that omitting it didn't lead to any build errors.  I'll place them back in the CMakeList files.

gavanderhoorn, line 12: Could be that you have Eigen installed in the default location (ie: `/usr/include`), which is always on the include path, or some other dependency already did a `find_package(Eigen ..)` somewhere.

jrgnicho:
Last commit reinstated find_package(Eigen ...) and squashed all previous commits into one.

gavanderhoorn:
Looking good. We don't want to add a `build_depend` on `Eigen` to the manifest? See earlier comment for the `rosdep` keys.

jrgnicho:
Would omitting it from the package.xml files break anything?

gavanderhoorn:
It would on systems with no `Eigen` installed. A `rosdep check ..` would not report `Eigen` not being installed then.

I don't really see a reason to not explicitly state all dependencies though.

jrgnicho:
build depends for Eigen have been added to the package.xml files

gavanderhoorn:
I think this is good to go.

+1 from me.

Issue #73, opened 2015-03-17
Cartesian Trajectory Point Unit Test Fails


Jmeyer1292:
The unit test covering the number of discretizations of a given sample space currently fails.

```
Value of: NUM_SAMPLED_ORIENT
  Actual: 216
Expected: solutions.size()
Which is: 124
/home/jon/ros/godel_ws/src/descartes/descartes_trajectory/test/cart_trajectory_pt.cpp:90: Failure
Value of: NUM_SAMPLED_ORIENT
  Actual: 216
Expected: joint_solutions.size()
Which is: 124
[ INFO] [1426606593.117965884]: Testing fuzzy both point
/home/jon/ros/godel_ws/src/descartes/descartes_trajectory/test/cart_trajectory_pt.cpp:94: Failure
Value of: NUM_SAMPLED_BOTH
  Actual: 287496
Expected: solutions.size()
Which is: 165044
/home/jon/ros/godel_ws/src/descartes/descartes_trajectory/test/cart_trajectory_pt.cpp:96: Failure
Value of: NUM_SAMPLED_BOTH
  Actual: 287496
Expected: joint_solutions.size()
Which is: 165044
[  FAILED  ] CartTrajPt.getPoses (233 ms)
```

Jmeyer1292:
The uniform method is correctly discretizing the sample space of the points it is given. The problem is that the CartesianRobot is filtering out points with the isValid(pose) method. This method reconstructs the XYZ rotations of the given pose object but it does not generate the exact same values as were used in the input.

The CartesianRobot tests if the rx, ry, rz values of the  rotation are within in [-0.5, 0.5] radians. These are frequently exceeded in the reconstructed points. 

After talking with @jrgnicho, I propose we simply make the cartesian robot's bounds larger for the purposes of this unit test.

PR #75, opened 2015-03-17
Trajectory Discretization Unit Tests (Joint Trajectory Point)

Jmeyer1292:
This pull request changes the settings of Cartesian Robot in a particular unit test to accept any joint values. Please see issue #73.

### DIFF
--- a/descartes_trajectory/test/cart_trajectory_pt.cpp
+++ b/descartes_trajectory/test/cart_trajectory_pt.cpp
@@ -77,7 +77,7 @@ TEST(CartTrajPt, getPoses)
   std::vector<std::vector<double> >joint_solutions;
 
   ROS_INFO_STREAM("Testing fuzzy pos point");
-  CartesianRobot robot(POS_TOL+2*EPSILON, ORIENT_TOL+2*EPSILON);
+  CartesianRobot robot(POS_TOL+2*EPSILON, M_PI*ORIENT_TOL+2*EPSILON);
   fuzzy_pos.getCartesianPoses(robot, solutions);
   EXPECT_EQ(solutions.size(), NUM_SAMPLED_POS);
   fuzzy_pos.getJointPoses(robot,joint_solutions);
--- a/descartes_trajectory/test/cart_trajectory_pt.cpp
+++ b/descartes_trajectory/test/cart_trajectory_pt.cpp
@@ -77,7 +77,7 @@ TEST(CartTrajPt, getPoses)
   std::vector<std::vector<double> >joint_solutions;
 
   ROS_INFO_STREAM("Testing fuzzy pos point");
-  CartesianRobot robot(POS_TOL+2*EPSILON, M_PI*ORIENT_TOL+2*EPSILON);
+  CartesianRobot robot(POS_TOL+2*EPSILON, 2*M_PI+2*EPSILON);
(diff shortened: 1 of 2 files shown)


descartes_trajectory/test/cart_trajectory_pt.cpp
```diff
@@ -77,7 +77,7 @@ TEST(CartTrajPt, getPoses)
   std::vector<std::vector<double> >joint_solutions;
 
   ROS_INFO_STREAM("Testing fuzzy pos point");
-  CartesianRobot robot(POS_TOL+2*EPSILON, ORIENT_TOL+2*EPSILON);
+  CartesianRobot robot(POS_TOL+2*EPSILON, M_PI*ORIENT_TOL+2*EPSILON);
```
jrgnicho: @Jmeyer1292  Would it not be more appropriate to set ORIENT_TOL to a larger value?

Jmeyer1292:
I went ahead and got rid of ORIENT_TOL in the constructor of the Cartesian Robot all-together. The important thing for now is that it accepts any joint values.

shaun-edwards:
@jrgnicho, please review.

descartes_trajectory/test/cart_trajectory_pt.cpp
```diff
@@ -77,7 +77,7 @@ TEST(CartTrajPt, getPoses)
   std::vector<std::vector<double> >joint_solutions;
 
   ROS_INFO_STREAM("Testing fuzzy pos point");
-  CartesianRobot robot(POS_TOL+2*EPSILON, ORIENT_TOL+2*EPSILON);
+  CartesianRobot robot(POS_TOL+2*EPSILON, 2*M_PI+2*EPSILON);
```
jrgnicho: Why not set the value of ORIENT_TOL to what it needs to be (2*M_PI apparently) instead of ignoring it?

shaun-edwards:
Good news guys...the build system caught this failure...see [here](http://jenkins.ros.org/job/devel-hydro-descartes/ARCH_PARAM=amd64,UBUNTU_PARAM=precise,label=devel/3/console).  Can we get this done?

descartes_trajectory/test/cart_trajectory_pt.cpp
Jmeyer1292: I was just trying to make the robot accept arbitrary angles without changing any other factors. ORIENT_TOL is used to calculate the number of discretizations. What do you think should be done to 'fix' it?

jrgnicho: I'd set **const double ORIENT_TOL = 2*M_PI**

Jmeyer1292: Well as long as you're okay with me changing the discretization size. Right now it'll take ~ 30 \* 30 \* 30 \* 10 \* 10 \* 10 = 27 million samples for a given point.

jrgnicho: Yes, that's a huge number of test.  Feel free to change that as well

shaun-edwards:
@Jmeyer1292, is this one good to go?

Jmeyer1292:
With this last small change, I believe @jrgnicho and myself are happy.

shaun-edwards:
+1...thanks for the contribution and the reviews.

PR #76, opened 2015-03-17
Axial symmetric pt

Jmeyer1292:
This pull request adds a sub-class of Cartesian Point that specializes it for the common case where one does not care about the yaw of the tool point as it moves through a path.

### DIFF
--- a/descartes_trajectory/src/axial_symmetric_pt.cpp
+++ b/descartes_trajectory/src/axial_symmetric_pt.cpp
@@ -1,18 +1,43 @@
 #include "descartes_trajectory/axial_symmetric_pt.h"
 
 
+using descartes_trajectory::TolerancedFrame;
+using descartes_trajectory::AxialSymmetricPt;
+using namespace descartes_core::utils;
+
+static TolerancedFrame makeRotationalAxis(AxialSymmetricPt::FreeAxis axis)
+{
+  using namespace descartes_trajectory;
+
+  Eigen::Affine3d rot = Eigen::Affine3d::Identity();
+  PositionTolerance pos_tol = ToleranceBase::zeroTolerance<PositionTolerance>(0,0,0);
+  OrientationTolerance orient_tol = ToleranceBase::createSymmetric<OrientationTolerance>(0.0, 0.0, 0.0, 
+                                    ((axis == AxialSymmetricPt::X_AXIS) ? 2*M_PI : 0.0),
+                                    ((axis == AxialSymmetricPt::Y_AXIS) ? 2*M_PI : 0.0),
+                                    ((axis == AxialSymmetricPt::Z_AXIS) ? 2*M_PI : 0.0));
+  return TolerancedFrame(rot, pos_tol, orient_tol);
+} 
+
+
 namespace descartes_trajectory
 {
 
-AxialSymmetricPt::AxialSymmetricPt(double x, double y, double z,
-                                   double rx, double ry, double rz,
-                                   double orient_increment) :
-  CartTrajectoryPt(TolerancedFrame(
-                     descartes_core::utils::toFrame(x, y, z, rx, ry, rz,
-                                                    descartes_core::utils::EulerConventions::XYZ),
-                     ToleranceBase::zeroTolerance<PositionTolerance>(x, y, z), // Fixed position
-                     // Fixed rx, ry, free rotation about z
-                     ToleranceBase::createSymmetric<OrientationTolerance>(rx, ry, 0, 0, 0, 2.0 * M_PI)),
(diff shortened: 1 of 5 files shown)


shaun-edwards:
Could this be implemented in a such a way that we could choose the redundant axis?  Say pass in an enumeration for the redundant axis?

Jmeyer1292:
Sure

shaun-edwards:
Also, there should be a test for this point.  I believe there are general tests for any trajectory point.  At the very least, this point should be added to these tests.

Jmeyer1292:
@shaun-edwards I can specify the redundant axis of rotation but for that to make sense, is it also fair to say I would need to give the user the option of choosing the rotation order (XYZ, ZYX, etc)?

Also what kind of tests do you envision for this point?

shaun-edwards:
Regarding testing, add it to the `TrajectoryPt` typed-test, [here](https://github.com/ros-industrial-consortium/descartes/blob/hydro-devel/descartes_trajectory/test/trajectory_pt.cpp#L74).  You should also turn the [CartesianPt](https://github.com/ros-industrial-consortium/descartes/blob/hydro-devel/descartes_trajectory/test/cart_trajectory_pt.cpp) test into a typed test and add it there as well.  Although it seems like there is a better way than to add it in two places.

As for rotation order, we could enforce that the all other rotation nominally be zero.  If a transformation is required, it could be specified as the `tool_base`, see [here](https://github.com/ros-industrial-consortium/descartes/blob/hydro-devel/descartes_trajectory/include/descartes_trajectory/cart_trajectory_pt.h#L292).

jrgnicho:
@Jmeyer1292 the orientation is assumed to be the result of rotations about fixed axis Z, Y and X in that particular order (post multiplication).  This assumption is made not only in the axially symmetric point but also in the **uniform()** method of the CartesianPt  as well.  Changing the order of rotation in the axially symmetric point will require that the **uniform()** method be modified to reflect this.

shaun-edwards:
@Jmeyer1292, ping.

Jmeyer1292:
@shaun-edwards I'll get back onto this asap. I promise I did not forget it.

Jmeyer1292:
@shaun-edwards  @jrgnicho Have a look when you get the chance. I added the ability to choose the axis of rotation of the nominal pose passed in. Seems to work correctly in simulation. Adding tests atm.

Jmeyer1292:
Unit tests added

descartes_trajectory/src/axial_symmetric_pt.cpp
```diff
@@ -0,0 +1,46 @@
...
+  CartTrajectoryPt(pose,
+                   makeRotationalAxis(axis),
+                   Eigen::Affine3d::Identity(),
+                   Eigen::Affine3d::Identity(),
+                   0.0, // The position discretization
+                   orient_increment) // Orientation discretization (starting at -2Pi and marching to 2Pi)
+{
+}
+
+} // end of namespace descartes_trajectory
```
jrgnicho, line 46: These calls to the **CartTrajectoryPt** constructor don't initialize the **wobj_pt_** ToleranceFrame with the right lower and upper bounds which are used to discretize the point by the [uniform](https://github.com/ros-industrial-consortium/descartes/blob/hydro-devel/descartes_trajectory/src/cart_trajectory_pt.cpp#L42) method.  See the amos version of the AxialSymmetricPt [here](https://github.com/swri-robotics/amos/blob/master/ubuntu/amos_core/include/amos_core/axial_symmetric_pt.h#L41)

Jmeyer1292, line 46: How so? The nominal transform is in wobj_base_ and the free axis rotation is represented by wobj_pt_.

jrgnicho, line 46: In the original version of the AxialSymetricPt the wobj_pt_ was initialized to the x,y,z,rx,ry,0 and wobj_base_ was set to Identity.  This resulted in fixed values for the translation and the rotations about x and y; only the rotation about z was discretized in the uniform method of the CartTrajectoryPt.  But, yes since the wobj_base now holds the nominal pose then the result should be equivalent, at least when rotating freely about z.  I think this new implementation is more flexible than the original one.

shaun-edwards:
Still missing [Trajectory Pt](https://github.com/ros-industrial-consortium/descartes/blob/hydro-devel/descartes_trajectory/test/trajectory_pt.cpp#L74) test.  Also, the Cartesian tests should be changed into typed tests.

Jmeyer1292:
@shaun-edwards I added the axial point to the trajectorypt type test, but I'm still a little fuzzy on converting the cartesian point tests to type tests. What is the shared test logic for cartesian and axial symmetric pt?

shaun-edwards:
+1

Issue #78, opened 2015-03-25
Shared pointer usage in Planners


Jmeyer1292:
The current interface (and implementations) for descartes path planners defines an initialize method which takes as an argument a reference to a shared pointer to a const robot model. This implies that the interface might make changes to the shared pointer itself. It also prevents users from initializing a shared pointer to a robot model and then simply passing it into the constructor (as that requires the compiler to make two levels of indirection to ConstPtr and then to ConstPtr&).

The fix is to simply take the shared pointer by value and use std::move (since we're in c++11 land now) to get it all the way down the chain to planning graph.

jrgnicho:
How about passing the robot model by const reference as done [here](https://github.com/ros-industrial-consortium/descartes/blob/hydro-devel/descartes_core/include/descartes_core/trajectory_pt.h#L107)?

Jmeyer1292:
The planners take a copy of the robot model pointer. I assume the reason that shared pointers are used in the first place is so that the lifespan of the model is tied to the lifespan of the planner. This change just makes it clear that the planner participates in ownership of the robot model and also allows users to pass shared pointers to their robot model without jumping through hoops.

shaun-edwards:
Just for some history...the original implementation might have used a raw pointer.  These were all changed to shared pointers in order to remove raw pointers from the public interface.  I wouldn't read too much into ownership implications.  My suggestion is to do what is right.  I would also prefer that this change only deprecate the existing method (this seems easy enough).

Jmeyer1292:
This change should not break any existing code.

PR #79, opened 2015-03-25
Issue #78 - Shared Pointer Usage Adjustments in Planners

Jmeyer1292:
This pull requests changes the path planner interface to take its robot model shared pointer by value instead of by reference. Other changes to implementations to facilitate this.

### DIFF
--- a/descartes_planner/src/dense_planner.cpp
+++ b/descartes_planner/src/dense_planner.cpp
@@ -31,10 +31,10 @@ DensePlanner::~DensePlanner()
 
 }
 
-bool DensePlanner::initialize(descartes_core::RobotModelConstPtr &model)
+bool DensePlanner::initialize(descartes_core::RobotModelConstPtr model)
 {
   planning_graph_ = boost::shared_ptr<descartes_planner::PlanningGraph>(
-      new descartes_planner::PlanningGraph(model));
+      new descartes_planner::PlanningGraph(std::move(model)));
   error_code_ = descartes_core::PlannerErrors::EMPTY_PATH;
   return true;
 }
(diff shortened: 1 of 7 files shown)


jrgnicho:
Sorry for not replying to the issue sooner, but see my comment [here](https://github.com/ros-industrial-consortium/descartes/issues/78)

descartes_planner/src/dense_planner.cpp
```diff
@@ -31,10 +31,10 @@ DensePlanner::~DensePlanner()
 
 }
 
-bool DensePlanner::initialize(descartes_core::RobotModelConstPtr &model)
+bool DensePlanner::initialize(descartes_core::RobotModelConstPtr model)
 {
   planning_graph_ = boost::shared_ptr<descartes_planner::PlanningGraph>(
-      new descartes_planner::PlanningGraph(model));
+      new descartes_planner::PlanningGraph(std::move(model)));
```
jrgnicho, line 37: Is this really necessary for passing a shared pointer?

Jmeyer1292, line 37: Moving the shared pointer avoids an atomic reference access and write. It also shows that we're just passing that value along. It's a small pro, but I think it is 'most' correct. Omitting the move wouldn't hurt the logic of the code.

jrgnicho, line 37: Would this not prevent incrementing the reference count of the shared pointer? As far as I understand the use of move semantics in this case transfers ownership of the object, so if the containing object goes out of scope could that invalidate other shared pointers?

Jmeyer1292, line 37: The shared pointer is "anchored" in main. That pointer is copied into the initialize method and then moved down to planning graph. When initialize is done the reference count on the pointer should be 2.

jrgnicho, line 37: +1

PR #80, opened 2015-03-25
Issue #78 shared pointer usage in planners

madratman:
"The fix is to simply take the shared pointer by value and use std::move (since we're in c++11 land now) to get it all the way down the chain to planning graph." Implemented in planning_graph, dense and sparse planners and their header files.

### DIFF
--- a/descartes_core/include/descartes_core/path_planner_base.h
+++ b/descartes_core/include/descartes_core/path_planner_base.h
@@ -51,7 +51,7 @@ public:
    * @param model robot model implementation for which to plan a path
    * @param config A map containing the parameter/value pairs.
    */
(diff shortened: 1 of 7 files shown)


madratman:
Okay. Just realized this has already been done by @Jmeyer1292

madratman:
I missed descartes_core/include/descartes_core/path_planner_base.h

jrgnicho:
This is already being addressed in PR #79 .  However, thank you for submitting a PR for this.

Jmeyer1292:
Yes, thank you for helping. If wish to help out, I'll be sure to post issues before I start working on the solution. :smiley:

madratman:
That would be great! Yes, I'd like to start contributing.

Issue #84, opened 2015-03-27
Incorrect discretization in the CartesianPt getCartesianPoses method


jrgnicho:
The discretization of the cartesian poses in the [getCartesianPoses()](https://github.com/ros-industrial-consortium/descartes/blob/hydro-devel/descartes_trajectory/src/cart_trajectory_pt.cpp#L187) method does not take into account the **tool_pt_** and **tool_base_** members when performing this computation.  The correct way is already implemented in the [getJointPoses()](https://github.com/ros-industrial-consortium/descartes/blob/hydro-devel/descartes_trajectory/src/cart_trajectory_pt.cpp#L300) method.

Jmeyer1292:
Addresed by #85

PR #85, opened 2015-03-27
Incorrect cartesian point discretization 

jrgnicho:
This is the fix for issue #84 
The discretization of the cartesian poses is now done in the **computeCartesianPoses** method which is invoked by the two method (getJointPoses and getCartesianPoses) that need this result.

### DIFF
--- a/descartes_trajectory/src/cart_trajectory_pt.cpp
+++ b/descartes_trajectory/src/cart_trajectory_pt.cpp
@@ -215,7 +215,7 @@ void CartTrajectoryPt::getCartesianPoses(const RobotModel &model, EigenSTL::vect
   if(computeCartesianPoses(all_poses))
   {
     poses.reserve(all_poses.size());
-    for(auto pose: all_poses)
+    for(const auto& pose: all_poses)
     {
       if(model.isValid(pose))
       {
@@ -337,7 +337,7 @@ void CartTrajectoryPt::getJointPoses(const RobotModel &model,
   if(computeCartesianPoses(poses))
   {
     poses.reserve(poses.size());
-    for(auto pose: poses)
+    for(const auto& pose: poses)
     {
       std::vector<std::vector<double> > local_joint_poses;
       if(model.getAllIK(pose, local_joint_poses))
--- a/descartes_trajectory/src/cart_trajectory_pt.cpp
+++ b/descartes_trajectory/src/cart_trajectory_pt.cpp
@@ -184,29 +184,60 @@ bool CartTrajectoryPt::getNominalCartPose(const std::vector<double> &seed_state,
   return true;  //TODO can this ever return false?
 }
 
-void CartTrajectoryPt::getCartesianPoses(const RobotModel &model, EigenSTL::vector_Affine3d &poses) const
+bool CartTrajectoryPt::computeCartesianPoses(EigenSTL::vector_Affine3d &poses) const
 {
-  poses.clear();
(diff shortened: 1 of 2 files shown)


shaun-edwards:
+1

descartes_trajectory/src/cart_trajectory_pt.cpp
```diff
@@ -184,29 +184,60 @@ bool CartTrajectoryPt::getNominalCartPose(const std::vector<double> &seed_state,
...
+
+void CartTrajectoryPt::getCartesianPoses(const RobotModel &model, EigenSTL::vector_Affine3d &poses) const
+{
+  EigenSTL::vector_Affine3d all_poses;
+  poses.clear();
+
+  if(computeCartesianPoses(all_poses))
+  {
+    poses.reserve(all_poses.size());
+    for(auto pose: all_poses)
```
Jmeyer1292: auto&, const auto&, or auto&& will avoid extra copies here.

descartes_trajectory/src/cart_trajectory_pt.cpp
```diff
@@ -184,29 +184,60 @@ bool CartTrajectoryPt::getNominalCartPose(const std::vector<double> &seed_state,
...
+    {
+      if(model.isValid(pose))
+      {
+        poses.push_back(pose);
```
Jmeyer1292, line 222: Since we're doing the C++11 thing, situations like this are great opportunities for using move semantics. Here transfering ownership of the object state is preferrable to copying. I don't know if eigen supports move constructors, but food for thought.

descartes_trajectory/src/cart_trajectory_pt.cpp
```diff
@@ -301,18 +332,13 @@ void CartTrajectoryPt::getJointPoses(const RobotModel &model,
...
-  size_t sample_size = sampled_wobj_pts.size() * sampled_tool_pts.size();
-  joint_poses.reserve(sample_size);
-  for(size_t wobj_pt = 0; wobj_pt < sampled_wobj_pts.size(); ++wobj_pt)
+
+  EigenSTL::vector_Affine3d poses;
+  if(computeCartesianPoses(poses))
   {
-    for(size_t tool_pt = 0; tool_pt < sampled_tool_pts.size(); ++tool_pt)
+    poses.reserve(poses.size());
+    for(auto pose: poses)
```
Jmeyer1292: See above.

descartes_trajectory/src/cart_trajectory_pt.cpp
jrgnicho, line 222: Where in the **getCartesianPoses** method could you apply move semantics?

descartes_trajectory/src/cart_trajectory_pt.cpp
jrgnicho: Yes, using references would  avoid making copies. +1

jrgnicho:
Changed auto to const auto&.  @Jmeyer1292 please review

descartes_trajectory/src/cart_trajectory_pt.cpp
Jmeyer1292, line 222: When you're pushing pose into the poses vector, you're essentially moving the data from the all_poses vector (you don't use it afterwards). So a std::move could wrap the pose object on line 222.

Jmeyer1292:
Looks nice. If you've tested and it works, then +1 from me.

Issue #86, opened 2015-03-27
Improvements to the getClosestJointPose method


jrgnicho:
In [CartesianPt::getClosestJointPose](https://github.com/ros-industrial-consortium/descartes/blob/hydro-devel/descartes_trajectory/src/cart_trajectory_pt.cpp#L212) ,FK is called with the seed pose and the resulting x,y,z, rx, ry,rz values are check against those in the point to find the closest cartesian pose.   Then IK is called for the closest cartesian pose in order to find the corresponding joints, however when IK fails the method does not return a joint pose.  This behavior seems lacking since in theory there should always be a **closest** joint pose.  A solution for this would consist of getting all the valid joint poses and for the point and comparing against the seed to find the closest pose in joint space.

PR #87, opened 2015-03-27
fix to getClosestJointPose

jrgnicho:
When IK for the closest pose fails then the search for the closest pose is done in joint space

### DIFF
--- a/descartes_trajectory/src/cart_trajectory_pt.cpp
+++ b/descartes_trajectory/src/cart_trajectory_pt.cpp
@@ -130,6 +130,27 @@ namespace descartes_trajectory
   return rtn;
 }
 
+double distance(const std::vector<double>& j1, const std::vector<double>& j2)
+{
+ double rt = 0;
+ double d;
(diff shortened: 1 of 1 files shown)


shaun-edwards:
This change could significant increase the complexity of the method call.  I agree with the rational in #86, but can we come up with a less complex implementation?  Also, this mixes distance metrics.  First cartesian distance is used, then joint distance is used.

jrgnicho:
An alternative would be to buffer up the poses in an array order by cartesian distance, in that case if IK fails on the closest then the next closest could be tested for IK and so on.  This approach would avoid mixing distance metrics but it would not guarantee a result.

jrgnicho:
While I agree that the complexity will increase when the search in cartesian space fails, in my testing of this change in the amos project it rarely resorted to a joint search between points that were very close.  I don't think that mixing distance metrics here is necessarily detrimental.

shaun-edwards:
+1

PR #92, opened 2015-03-31
adds descartes metapackage

madratman:


shaun-edwards:
+1...thanks for the contribution.

PR #96, opened 2015-04-01
Issue#77 uuid replacement

Jmeyer1292:
This PR creates a trajectory ID class that encapsulates a unique state object. I tried to design the implementation so it will be easy to switch the underlying state object to any type we want (including UUIDs). 

The current system works by keeping an atomic integer and incrementing it every time we generate a new point. This is very fast compared to making UUIDs and we save something like 90% of the time previously required. It has limitations though:
1. We can only generate 2^64 - 1 unique IDs in a single process. 
2. IDs are only unique to a given process.

There might be more limitations and I invite conversation on what the right answer is. I find the performance improvement compelling and was my primary motivation for doing this. Trajectories of several hundred points solve more or less instantly for me now.

### DIFF
--- a/descartes_core/CMakeLists.txt
+++ b/descartes_core/CMakeLists.txt
@@ -13,6 +13,8 @@ find_package(Eigen REQUIRED)
 catkin_package(
   INCLUDE_DIRS
     include
+  LIBRARIES
+    descartes_core
   CATKIN_DEPENDS
     roscpp
     moveit_core
@@ -31,4 +33,13 @@ include_directories(include
                     ${Boost_INCLUDE_DIRS}
                     ${Eigen_INCLUDE_DIRS}
 )
+
+add_library(descartes_core
+            src/trajectory_id.cpp
+)
+
+target_link_libraries(descartes_core
+                      ${catkin_LIBRARIES}
+)
+
 
--- a/descartes_core/include/descartes_core/trajectory_id.h
+++ b/descartes_core/include/descartes_core/trajectory_id.h
@@ -1,0 +1,102 @@
+#ifndef TRAJECTORY_ID_H
+#define TRAJECTORY_ID_H
+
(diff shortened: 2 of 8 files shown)


shaun-edwards:
+1...while this is less powerful than the UUID, I don't see a compelling case that requires UUID.  The original reason for using UUID was convenience.  I'll keep this open for a few days to allow others to comment.

ClayFlannigan:
As a potential enhancement, what do you think about initializing the ID with a call to time() or clock() or a pseudo random generator?  As is, this is guaranteed to collide, although I'm not sure of the use cases where it would matter.

shaun-edwards:
Depending on the time resolution and how time changes are handled or the random number generator specifics, there isn't a guarantee of uniqueness within or between processes.  UUIDs provide this guarantee but apparently at great expense.  The current approach can guarantee uniqueness  within a process, which is required at a minimum.

ClayFlannigan:
'm proposing to initialize with a pseudo random number which would still be guaranteed unique within a process since they are serially incremented after the initialization.  Between processes you would have a good chance of being unique, but if/when it fails, the result could be hard to detect/debug.  So maybe it is better to just do the simple thing and force the collisions to happen.

Jmeyer1292:
If we generated 1 million ids per second, we could operate for the next 2^64(1e6 \* 60 \* 60 \* 24 \* 365) =  584942.4 years before we overrun.

shaun-edwards:
Sounds like @Jmeyer1292 thinks there should be a check for overflow in the ID generation method. ;)

ClayFlannigan:
The nearest star, Proxima Centauri, is 4.01126E+19 millimeters from Earth.  If at 1 mm discretization, we will overflow twice before we finish the plan.  I recommend a 128 bit integer.

PR #98, opened 2015-04-03
Small Changes to Fix Build Failures

Jmeyer1292:
This PR addresses two issues causing the build of Descartes to fail:
1. There appears to be a non-ascii character immediately after the meta-package tag in the descartes package.xml.
2. I missed some things when rebasing code form a previous pull-request (from the uuid change).

Sorry!


shaun-edwards:
+1...Thanks for catching this.

PR #102, opened 2015-04-06
Doxygen Mainpage Updates

Jmeyer1292:
This PR updates the mainpage.dox of descartes_trajectory to include the axially symmetric point class.


shaun-edwards:
+1.

Issue #103, opened 2015-04-06
TrajectoryPt cloneTo() method unimplemented


Jmeyer1292:
The base class for TrajectoryPt defines a virtual method cloneTo:

``` c++
virtual void cloneTo(TrajectoryPt &clone) const
  {
    clone = *this;
    clone.setID(TrajectoryID::make_id());
  }
```

This method is unimplemented in all of the child classes and it will not do the 'expected' thing. What I believe it will do is 'slice' the 'this' variable and only copy the information that it is aware of in the base class itself. 

The fix is probably to create a clone function that returns an entirely new TrajectoryPtPtr with the appropriate underlying type. The implementation will probably look something like:

``` c++
virtual TrajectoyPtPtr clone() const
{
// add cloning info for timing and ids
return new SomeTypeOfPt(*this);
}
```

Or it could by a cloneTo method. The difficulty there is memory management. You'd probably have to take a shared pointer by reference and call it's reset function.

shaun-edwards:
@jrgnicho...any ideas?

jrgnicho:
I'm on board with requiring each subclass to override the "clone()" method that returns a TrajectoryPtPtr that has been initialized with the an object of that particular class.

Jmeyer1292:
#105 addresses this issue

PR #105, opened 2015-04-09
Issue #103 - polymorphic clone and copy

Jmeyer1292:
Issue #103: This PR changes the trajectory point base class to have a pure virtual function copy() that returns a shared pointer to a trajectory point. It re-factors the clone() method to use this copy routine. 

Tests have been similarly updated. This PR introduces the requirement that all derivatives of trajectory pt must implement a copy function.


shaun-edwards:
+1

PR #113, opened 2015-04-16
MoveitStateAdaptor getIK joint limit check

jrgnicho:
This PR modifies the **getIK** method in order to verify that the joint solution is within the joint limits as defined in the **URDF** file for the robot model.

### DIFF
--- a/descartes_moveit/src/moveit_state_adapter.cpp
+++ b/descartes_moveit/src/moveit_state_adapter.cpp
@@ -90,8 +90,8 @@ bool MoveitStateAdapter::initialize(const std::string& robot_description, const 
 
   if (seed_states_.empty())
   {
-    ROS_INFO_STREAM("Generated random seeds");
     seed_states_ = seed::findRandomSeeds(*robot_state_, group_name_, SAMPLE_ITERATIONS);
+    ROS_INFO_STREAM("Generated "<<seed_states_.size()<< " random seeds");
   }
 
   const moveit::core::JointModelGroup* joint_model_group_ptr = robot_state_->getJointModelGroup(group_name);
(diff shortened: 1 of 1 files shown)


jrgnicho:
@shaun-edwards please review

descartes_moveit/src/moveit_state_adapter.cpp
```diff
@@ -145,9 +145,9 @@ bool MoveitStateAdapter::getIK(const Eigen::Affine3d &pose, std::vector<double>
                               tool_frame_))
   {
     robot_state_->copyJointGroupPositions(group_name_, joint_pose);
-    if(isInCollision(joint_pose))
+    if(!isValid(joint_pose))
```
shaun-edwards, line 148: Does `isValid` check for collisions?

descartes_moveit/src/moveit_state_adapter.cpp
```diff
@@ -109,7 +109,7 @@ bool MoveitStateAdapter::initialize(const std::string& robot_description, const
     if (world_frame_ != robot_state_->getRobotModel()->getModelFrame())
     {
       logWarn("World frame '%s' does not match model root frame '%s', all poses will be transformed to world frame '%s'",
-               world_frame_.c_str(), link_names.front().c_str(),world_frame_.c_str());
+               world_frame_.c_str(), robot_state_->getRobotModel()->getModelFrame().c_str(),world_frame_.c_str());
```
shaun-edwards, line 112: Just a suggestion, but if we change log messages, we should also take the opportunity to update to the ROS macros.

descartes_moveit/src/moveit_state_adapter.cpp
shaun-edwards, line 148: Yep...it appears it does ;)

shaun-edwards:
+1...you can merge.

jrgnicho:
It does check for collisions.  I'll update the macros.  Thanks

jrgnicho:
I replaced the log messages for the equivalent ROS macros but started getting lots of warnings during the build.  It looks like this might take more than a simple replacement so I created an issue for it here #116.  I'll address that in a different PR.

Issue #110, opened 2015-04-09
Cartesian Point's getClosestJointPose() doesn't account for all transforms


Jmeyer1292:
The getClosestJointPose function of the cartesian trajectory point does FK on a set of joint values and then checks to see if that solution is within the tolerances of the toleranced-frame "wobj_pt_".

Cartesian point, however, specifies 4 possible frames to fully define the position of the robot end-effector and this method only takes into account one of them. If someone specifies the other transformations as other than identity (say they use AxialSymmetricPt) then this method will fail.

I'm not sure what the correct math is, but we need to find a way to compute the over-all tolerances in the world frame based on the four transformations (two of which are nominal and two of which are toleranced).

See: https://github.com/ros-industrial-consortium/descartes/blob/hydro-devel/descartes_trajectory/src/cart_trajectory_pt.cpp#L276-L289

shaun-edwards:
To make this problem tractable, we can assume that only one of the two toleranced frames is every defined as such.  The other toleranced frame will be defined as identify.  This may required re-defining constructors to ensure that this is the case.  While this may limit the general purpose of the CartTrajectoryPt, it's hard to identify a use case where two toleranced frames are required.  The original intent was only to provide a way of defining the frames in either reference frame (i.e. whichever was convenient).

Issue #114, opened 2015-04-16
AxialSymetricPt does not properly initialize tolerance frame in CartesianPt parent class


jrgnicho:
The AxialSymetricPt constructors assigns the nominal pose data to the wobj_base_member instead of the wobj_pt_ tolerance frame member.  Thus the correct translation and rotation bounds of the point aren't captured within the tolerance frame and they are set to zero instead.  This causes failures when calling the **getClosestJointPose()** method which in turn causes the SparsePlanner to fail at finding a path.

Jmeyer1292:
The issue is not with AxiallySymmetricPt. The issue (see #110) is that getClosestJointPose doesn't work with the 4 transforms  that can be specified in Cartesian point. Anyone who specifies more than just wob_pt_ will run into issues.

jrgnicho:
I agree that the getClosestJointPose() method is incomplete (see issue #43).  The proper solution would require using all transforms and evaluating the rotational bounds in quaternion space in order to avoid dealing with discontinuities inherent in the cartesian space euler angles.

PR #115, opened 2015-04-16
issue #114 AxialSymetricPt initialization

jrgnicho:
Related to issue #114.
This PR changes the AxialSymetricPt constructors so that they call the CartesianPt constructor that sets the nominal values of the tolerance frame. This allows for the translation and rotation bounds of the wobj_pt tolerance frame to be properly defined.  This change fixes the SparsePlanner which now gets valid joint poses whenever it calls the  **getClosestJointPose** method

### DIFF
--- a/descartes_trajectory/src/axial_symmetric_pt.cpp
+++ b/descartes_trajectory/src/axial_symmetric_pt.cpp
@@ -5,18 +5,46 @@ using descartes_trajectory::TolerancedFrame;
 using descartes_trajectory::AxialSymmetricPt;
 using namespace descartes_core::utils;
 
-static TolerancedFrame makeRotationalAxis(AxialSymmetricPt::FreeAxis axis)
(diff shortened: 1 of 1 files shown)


Jmeyer1292:
This is at best a work-around to the problem (see #110 and discussion in #114). The reason I used multiple frames was to allow the user to specify the axis of rotation of a nominal pose and have the point produce the expected behavior despite the fact we use euler-XYZ internally. This change, I believe, will result in discretizations that happen in the intermediate steps of the rotation, causing unexpected behavior.

jrgnicho:
Indeed, the changes here affect how the point is discretized but the resulting number of discretized poses and their values are the same.

shaun-edwards:
+1...We are going to accept this as a work-around for now.  This will not close #110.  A comment has been added to #110 regarding the proper fix.

PR #119, opened 2015-04-22
returns true when a solution is found

jrgnicho:
A "false" boolean was being returned by the the getClosestJointPose(...) method even though a solution was found.  This PR fixes this.


shaun-edwards:
+1...merging.

PR #125, opened 2015-06-22
Fix to Sparse Planner crash

jrgnicho:
After the introduction of the timing information in each trajectory point #65 the SparsePlanner crashes, possibly running into a segfault.  This error was encountered when planning for the robot setup in the [descartes_demo](https://github.com/jrgnicho/demo_descartes) package for which the SparsePlanner had previously worked.  This PR fixes this issue.

### DIFF
--- a/descartes_planner/src/sparse_planner.cpp
+++ b/descartes_planner/src/sparse_planner.cpp
@@ -763,11 +763,12 @@ int SparsePlanner::interpolateSparseTrajectory(const SolutionArray& sparse_solut
           const JointTrajectoryPt& last_joint_pt = joint_points_map_.at(cart_points_[pos-1]->getID());
           std::vector<double> last_joint_pose;
           last_joint_pt.getNominalJointPose(std::vector<double>(), *robot_model, last_joint_pose);
-          // look up dt
-          double dt = timing_cache_[pos].upper;
-          // check validity of joint motion
 
-          if (dt > 0.0 && !robot_model->isValidMove(last_joint_pose, aprox_interp, dt))
+          // retreiving timing constraint
+          const descartes_core::TimingConstraint& tm = timing_cache_[pos];
+
+          // check validity of joint motion
+          if (tm.isSpecified() && !robot_model->isValidMove(last_joint_pose, aprox_interp, tm.upper))
           {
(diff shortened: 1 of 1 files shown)


descartes_planner/src/sparse_planner.cpp
```diff
@@ -767,9 +767,11 @@ int SparsePlanner::interpolateSparseTrajectory(const SolutionArray& sparse_solut
           double dt = timing_cache_[pos].upper;
```
Jmeyer1292: I would suggest that we pull the entire TimingConstraint struct here.

descartes_planner/src/sparse_planner.cpp
```diff
@@ -767,9 +767,11 @@ int SparsePlanner::interpolateSparseTrajectory(const SolutionArray& sparse_solut
...
           // check validity of joint motion
 
-          if (!robot_model->isValidMove(last_joint_pose, aprox_interp, dt))
+          if (dt > 0.0 && !robot_model->isValidMove(last_joint_pose, aprox_interp, dt))
```
Jmeyer1292: ... And use `isSpecified()` here.

descartes_planner/src/sparse_planner.cpp
jrgnicho: Not sure that I follow, can you elaborate more on that?

Jmeyer1292:
``` c++
const descartes_core::TimingConstraint& tm = timing_cache_[pos];
```

followed by

``` c++
if (tm.isSpecified() && !robot_model->isValidMove(last_joint_pose, aprox_interp, tm.upper))
```

So that we're using the API for checking if the time is specified and not looking at the raw data. That's all.

Jmeyer1292:
Do you understand why the program is crashing when it returns a `replan` on a point that has timing associated with it? 

Anyway, after this change, +1 from me.

descartes_planner/src/sparse_planner.cpp
jrgnicho: Thanks for clarifying

jrgnicho:
@Jmeyer1292 there were two indices that were not being updated when interpolation failed and a replan was requested.
The latest commit contains the changes that you suggested.

Jmeyer1292:
Thanks for looking at this. +1

PR #126, opened 2015-06-23
JointTrajectoryPt isValid() Fix

Jmeyer1292:
This PR fixes a bug in Joint Trajectory Point's isValid() function that was causing it to check incorrect joint configurations. 

JointTrajectoryPt has, for every joint, a 'joint position' that consists of a nominal value and an upper and lower tolerance from that point. The latter are stored as displacements from the nominal, not as absolute positions themselves. The isValid function was treating them as absolute positions and thus was checking the wrong joint angles.

### DIFF
--- a/descartes_trajectory/include/descartes_trajectory/joint_trajectory_pt.h
+++ b/descartes_trajectory/include/descartes_trajectory/joint_trajectory_pt.h
@@ -31,43 +31,25 @@
 namespace descartes_trajectory
 {
 
-//TODO add warning if non-zero tolerances are specified because initial implementation will only allow fixed joints
-struct JointTolerance
-{
-  JointTolerance(): lower(0.), upper(0.) {};
-  JointTolerance(double low, double high):
-    lower(std::abs(low)), upper(std::abs(high)) {};
-
-  double lower, upper;
-};
-
+/**
+ * @brief Structure to specify a valid joint range [lower, upper] with a nominal position given by 'nominal'.
+ */
 struct TolerancedJointValue
 {
-  TolerancedJointValue() {};
-  TolerancedJointValue(double _nominal, double _tol_above, double _tol_below):
-    nominal(_nominal), tolerance(_tol_above, _tol_below) {};
-  TolerancedJointValue(double _nominal)
-  {
-    *this = TolerancedJointValue(_nominal, 0., 0.);
-  }
+  TolerancedJointValue(double nominal, double lower, double upper)
+    : nominal(nominal), lower(lower), upper(upper)
+  {}
 
-  double upperBound() const
-  {
-    return nominal+tolerance.upper;
-  }
-
-  double lowerBound() const
-  {
-    return nominal-tolerance.lower;
-  }
+  TolerancedJointValue(double nominal)
+    : nominal(nominal), lower(nominal), upper(nominal)
+  {}
 
   double range() const
   {
-    return upperBound() - lowerBound();
+    return upper - lower;
   }
 
-  double nominal;
-  JointTolerance tolerance;
+  double nominal, lower, upper;
 };
 
 /**@brief Joint Trajectory Point used to describe a joint goal for a robot trajectory.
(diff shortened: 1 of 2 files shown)


shaun-edwards:
Doesn't the Cartesian point consider the upper/lower as nominal values? See [here](https://github.com/ros-industrial-consortium/descartes/blob/hydro-devel/descartes_trajectory/include/descartes_trajectory/cart_trajectory_pt.h#L127).  Should they be the same.

Jmeyer1292:
@shaun-edwards I absolutely agree. In the course of fixing this issue, I took the opportunity to change the way the data was stored in the class itself to improve efficiency and usability.

I kept the same interface, so one can still initialize the point with TolerancedJointValues but the data is now stored in seperate vectors for the nominal, lower, and upper bounds. This simplifies the implementation of many of the class methods by quite a bit because there is no need to transform data into vectors.

Furthermore it means that users who have a JointTrajectoryPt can just call `nominal()` to get a reference to the nominal joint values. This is more user-friendly and efficient than making an indirect call through getNominalJointPosition. When I updated PlanningGraph to use this function internally in the inner-most loop of edge calculations, it doubled the speed edges were computed (by avoiding allocations).

descartes_trajectory/include/descartes_trajectory/joint_trajectory_pt.h
```diff
@@ -31,34 +31,24 @@
...
+  {}
+
+  TolerancedJointValue(double nominal)
+    : nominal(nominal), lower(nominal), upper(nominal)
+  {}
 
   double upperBound() const
   {
-    return nominal+tolerance.upper;
+    return upper;
```
jrgnicho: There is probably no need to have these accessor methods anymore since all the do is return a public member.

descartes_trajectory/include/descartes_trajectory/joint_trajectory_pt.h
```diff
@@ -31,34 +31,24 @@
...
   }
 
   double lowerBound() const
```
jrgnicho: Same as above

descartes_trajectory/include/descartes_trajectory/joint_trajectory_pt.h
```diff
@@ -31,34 +31,24 @@
...
   {
-    return nominal-tolerance.lower;
+    return lower;
   }
 
   double range() const
```
jrgnicho, line 47: It looks like this method is no longer being used since now the range is computed from the bound arrays in the JointTrajectoryPoint

descartes_trajectory/include/descartes_trajectory/joint_trajectory_pt.h
```diff
@@ -31,34 +31,24 @@
...
 struct TolerancedJointValue
 {
-  TolerancedJointValue() {};
-  TolerancedJointValue(double _nominal, double _tol_above, double _tol_below):
-    nominal(_nominal), tolerance(_tol_above, _tol_below) {};
-  TolerancedJointValue(double _nominal)
-  {
-    *this = TolerancedJointValue(_nominal, 0., 0.);
-  }
+  TolerancedJointValue(double nominal, double lower, double upper)
```
jrgnicho, line 39: It be good to document this constructor so that the **lower** and **upper** arguments are used as intended.

jrgnicho:
I like the performance gains introduced by these changes.  I makes me wonder if we should make the Descartes planners return  JointTrajectoryPt arrays so that we can take easily retrieve the joint values with the newly added **nominal()** method instead of the **getNominalJointPose()** method;  which was regarded as confusing by several people during the last ROS-I training.

descartes_trajectory/include/descartes_trajectory/joint_trajectory_pt.h
Jmeyer1292: +1

descartes_trajectory/include/descartes_trajectory/joint_trajectory_pt.h
Jmeyer1292, line 47: +1

descartes_trajectory/include/descartes_trajectory/joint_trajectory_pt.h
Jmeyer1292, line 39: +1

Jmeyer1292:
I agree. I intend to do that next.

I've updated the PR to incorporate your change requests. Everything still compile so I guess no one was using those functions.

jrgnicho:
Looks good, +1 from me.

PR #127, opened 2015-06-29
Updated the linearWeight function to use the new 'nominal' function in JointTrajectoryPt

Jmeyer1292:
This reduces copying by a huge amount, allowing the graph to build twice as fast for edge heavy graphs. The diff below looks kinda gnarly but all I did was remove the outer-most if check and the cruft related to calling `getJointPoses`.

### DIFF
--- a/descartes_planner/src/planning_graph.cpp
+++ b/descartes_planner/src/planning_graph.cpp
@@ -996,48 +996,33 @@ PlanningGraph::LinearWeightResult PlanningGraph::linearWeight(const JointTraject
   LinearWeightResult result;
   result.first = false;
 
-  // Retrieve the nominal joint position
-  std::vector<std::vector<double> > joint_poses_start;
-  start.getJointPoses(*robot_model_, joint_poses_start);
-
-  std::vector<std::vector<double> > joint_poses_end;
-  end.getJointPoses(*robot_model_, joint_poses_end);
-
-  // each should only return one
-  if (joint_poses_start.size() == 1 && joint_poses_end.size() == 1)
+  const std::vector<double>& start_vector = start.nominal();
+  const std::vector<double>& end_vector = end.nominal();
+  if (start_vector.size() == end_vector.size())
   {
-    const std::vector<double>& start_vector = joint_poses_start[0];
-    const std::vector<double>& end_vector = joint_poses_end[0];
-    if (start_vector.size() == end_vector.size())
+    // Check to see if time is specified and if so, check to see if the
+    // joint motion is possible in the window provided
+    if (end.getTiming().isSpecified() &&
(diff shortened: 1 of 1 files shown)


descartes_planner/src/planning_graph.cpp
```diff
@@ -996,48 +996,33 @@ PlanningGraph::LinearWeightResult PlanningGraph::linearWeight(const JointTraject
...
-      double vector_diff = 0;
-      for (unsigned i = 0; i < start_vector.size(); i++)
-      {
-        double joint_diff = std::abs(end_vector[i] - start_vector[i]);
-        vector_diff += joint_diff ;
-      }
-
-      result.first = true;
-      result.second = vector_diff;
       return result;
```
jrgnicho, line 1008: Nothing is being assigned to the **result** object any more, is that the intended behaviour?

descartes_planner/src/planning_graph.cpp
```diff
@@ -996,48 +996,33 @@ PlanningGraph::LinearWeightResult PlanningGraph::linearWeight(const JointTraject
...
-    const std::vector<double>& end_vector = joint_poses_end[0];
-    if (start_vector.size() == end_vector.size())
+  const std::vector<double>& start_vector = start.nominal();
+  const std::vector<double>& end_vector = end.nominal();
+  if (start_vector.size() == end_vector.size())
+  {
+    // Check to see if time is specified and if so, check to see if the
+    // joint motion is possible in the window provided
+    if (end.getTiming().isSpecified() &&
+        !robot_model_->isValidMove(start_vector, end_vector, end.getTiming().upper))
```
jrgnicho, line 1006: This check feels non-intuitive to me.   Why is the returned bool from isValidMove() being negated?

descartes_planner/src/planning_graph.cpp
Jmeyer1292, line 1008: result.first is set to false at the top of the function. If you look at the annotation-free version you'll see it still makes sense.

jrgnicho, line 1008: +1

descartes_planner/src/planning_graph.cpp
Jmeyer1292, line 1006: If the timing is specified and the move is NOT valid, then we return a false immediately.

jrgnicho, line 1006: +1

jrgnicho:
Did you by any chance validate these changes by running them on a simulated robot?

Jmeyer1292:
I did. I tested it on my SIA20D setup for Descartes.

jrgnicho:
Great.  +1 from me, feel free to merge

Issue #138, opened 2015-12-02
SparsePlanner Exceeds Max Planning Attempts


Jmeyer1292:
For trajectories and robot kinematics that can not easily be interpolated, the SparsePlanner may have to sample most or all of the original trajectory. Currently the planner will fail if it has to add more points than half the size of the input trajectory.

This should be changed. The SparsePlanner should add all of the points and return an answer if there is one.

Issue #137, opened 2015-12-02
SparsePlanner Does Not Fail on Unachievable Timing Constraints


Jmeyer1292:
Per unit tests added in #135, SparsePlanner does not fail when joint velocities in robot model are exceeded (and `isValidMove` fails).

Issue #136, opened 2015-12-02
SparsePlanner Modifies Input Trajectories


Jmeyer1292:
The existing implementation of `SparsePlanner` modifies the timing constraints of an input trajectory by setting them all to unconstrained. This side effect is neither obvious nor documented and may seriously surprise users.

See [the code](https://github.com/ros-industrial-consortium/descartes/blob/indigo-devel/descartes_planner/src/sparse_planner.cpp#L152).

PR #140, opened 2015-12-11
Sparse Planner Refactoring (Issues 136-138)

Jmeyer1292:
This PR address #136, #137, and #138. 

It makes the sparse planner sample both the positions and timing values of the input trajectory. For example, after sampling the first and last point in a three point trajectory with 1 second spacing, the final point will have a 2 second spacing. This is what we've done.

This PR also removes the maximum planning attempts count, and adds a test trajectory (a zig-zag pattern) that forces sparse planner to frequently replan with the test robot model.

### DIFF
--- a/descartes_planner/src/sparse_planner.cpp
+++ b/descartes_planner/src/sparse_planner.cpp
@@ -635,7 +635,7 @@ bool SparsePlanner::plan()
   // solving coarse trajectory
   bool replan = true;
   bool succeeded = false;
-  int max_replanning_attempts = cart_points_.size()/2;
+  int max_replanning_attempts = cart_points_.size();
   int replanning_attempts = 0;
   while(replan && getSparseSolutionArray(sparse_solution_array_))
   {
--- a/descartes_planner/src/sparse_planner.cpp
+++ b/descartes_planner/src/sparse_planner.cpp
@@ -30,18 +30,39 @@ using namespace descartes_trajectory;
 
 namespace
 {
-    std::vector<descartes_core::TimingConstraint>
-    extractTiming(const std::vector<descartes_core::TrajectoryPtPtr>& points)
+  descartes_core::TimingConstraint
+  cumulativeTimingBetween(const std::vector<descartes_core::TrajectoryPtPtr>& dense_points,
+                          size_t start_index,
+                          size_t end_index)
+  {
+    descartes_core::TimingConstraint tm (0.0, 0.0);
+    for (size_t i = start_index + 1; i <= end_index; ++i)
     {
-      std::vector<descartes_core::TimingConstraint> result;
-      result.reserve(points.size());
-
-      for (const auto& ptr : points)
+      const descartes_core::TimingConstraint& pt_tm = dense_points[i]->getTiming();
+      if (pt_tm.isSpecified())
       {
-        result.push_back(ptr->getTiming());
+        // Add time as normal
+        tm.upper += pt_tm.upper;
+        tm.lower += pt_tm.lower;
       }
-      return result;
+      else
+      {
+        // A single unspecified timing makes the range unspecified
+        tm = descartes_core::TimingConstraint();
+        break;
+      }
     }
+    return tm;
+  }
+
+  descartes_core::TrajectoryPtPtr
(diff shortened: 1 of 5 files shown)


descartes_planner/src/sparse_planner.cpp
```diff
@@ -30,18 +30,39 @@ using namespace descartes_trajectory;
...
+        tm = descartes_core::TimingConstraint();
+        break;
       }
-      return result;
     }
+    return tm;
+  }
+
+  descartes_core::TrajectoryPtPtr
+  copyAndSetTiming(const descartes_core::TrajectoryPtPtr& ptr, const descartes_core::TimingConstraint& tm)
```
shaun-edwards: Should this method be supported at a TrajectoryPt level?

descartes_planner/src/sparse_planner.cpp
```diff
@@ -589,9 +601,26 @@ void SparsePlanner::sampleTrajectory(double sampling,const std::vector<Trajector
...
+  
+  if (dense_trajectory_array.empty()) return;
+
+  // Add the first point
+  sparse_trajectory_array.push_back(dense_trajectory_array.front());
+  ss << "0 ";
+  // The first point requires no special timing adjustment
+  
+  int i; // We keep i outside of the loop so we can examine it on the last step
+  for (i = skip; i < dense_trajectory_array.size(); i+=skip)
```
shaun-edwards, line 604: Where is `skip` defined?

shaun-edwards:
Two general questions:
1. Do these changes affect the complexity of the Sparse Planner and performance in general?
2. Should we add tests that verify more complex timing constraints?

descartes_planner/src/sparse_planner.cpp
```diff
@@ -648,29 +686,53 @@ bool SparsePlanner::plan()
...
+            int prev_dense_id = point_pos;
+            int next_dense_id = std::get<0>(sparse_solution_array_[sparse_index]);
+            descartes_core::TimingConstraint tm = cumulativeTimingBetween(cart_points_, prev_dense_id, next_dense_id);
+            descartes_core::TrajectoryPtPtr copy_pt = copyAndSetTiming(cart_points_[next_dense_id], tm);
+
+            if (!planning_graph_->modifyTrajectory(copy_pt))
+            {
+              ROS_ERROR_STREAM("Could not modify trajectory point with id: " << copy_pt->getID());
+              replan = false;
+              succeeded = false;
```
jrgnicho, line 713: Should this condition break out of the while loop and return a failure?

descartes_planner/src/sparse_planner.cpp
Jmeyer1292, line 604: Skip is defined on line 601 of the new file to be:

``` c++
int skip = std::ceil(double(1.0f)/sampling);
```

This is unchanged from the original implementation.

descartes_planner/src/sparse_planner.cpp
Jmeyer1292, line 713: +1 If we run into this situation then we're modifying a point in the sparse trajectory that is NOT in the planning_graph. This should never happen, but sanity checks are good.

descartes_planner/src/sparse_planner.cpp
Jmeyer1292: While I hesitate to add too many additional methods to trajectory pt, I think this one might be common enough to be worth it.

I toyed with the notion of having a new copy() function that takes a timing value as an arg:

``` c++
auto copy = original_pt->copy(TimingContraint(1.0));
```

While I like the brevity of this, I think it's better to be explicit. +1

Jmeyer1292:
@shaun-edwards To your questions:
1. I do not believe that these changes will have a significant impact on the run time of the sparse planner. The current implementation adds a worst case `O(n)` calculation to compute the cumulative time between points each time a new (non-interpolated) point is added to the sparse plan. I don't have any hard numbers to back me up though.
2. It certainly can't hurt to do more testing. Do you have any ideas about what those might look like? I'd also like to add some performance benchmarks and regression tests per our earlier discussions.

shaun-edwards:
+1

PR #144, opened 2016-02-29
ROS Utilities

Jmeyer1292:
This PR introduces a new package `descartes_utilities` which currently exposes one file, `ros_conversions.h` that can be used to convert a Descartes trajectory to a standard vector of ROS JointTrajectoryPoints.

This is an extremely common task, and I find the same code sprinkled through every project that integrates Descartes with a ROS interface. In the future, I imagine this package as a home for other useful utilities for inputting and outputting trajectories through the Descartes motion planner.

### DIFF
--- a/descartes_utilities/CMakeLists.txt
+++ b/descartes_utilities/CMakeLists.txt
@@ -1,0 +1,38 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(descartes_utilities)
+
+find_package(catkin REQUIRED COMPONENTS
+  descartes_core
+  descartes_trajectory
+  trajectory_msgs
+)
+
+catkin_package(
+  INCLUDE_DIRS include
+  LIBRARIES descartes_utilities
+  CATKIN_DEPENDS descartes_core descartes_trajectory trajectory_msgs
+)
+
+include_directories(
+  include
+  ${catkin_INCLUDE_DIRS}
+)
+
+add_library(${PROJECT_NAME}
+  src/ros_conversions.cpp
+)
+
+add_dependencies(${PROJECT_NAME} 
+  ${catkin_EXPORTED_TARGETS}
+)
+
+install(TARGETS ${PROJECT_NAME}
+  ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+  LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+  RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
+)
+
+install(DIRECTORY include/${PROJECT_NAME}/
+  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
+  FILES_MATCHING PATTERN "*.h"
+)
--- a/descartes_utilities/include/descartes_utilities/ros_conversions.h
+++ b/descartes_utilities/include/descartes_utilities/ros_conversions.h
@@ -1,0 +1,51 @@
+/*
+ * Software License Agreement (Apache License)
+ *
+ * Copyright (c) 2016, Southwest Research Institute
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/*
+ *  ros_conversions.h
+ *
+ *  Created on: Feb 28, 2016
+ *  Author: Jonathan Meyer
+ */
+
+#ifndef DESCARTES_ROS_CONVERSIONS_H
+#define DESCARTES_ROS_CONVERSIONS_H
+
+#include <trajectory_msgs/JointTrajectory.h>
+#include <descartes_core/trajectory_pt.h>
+
+namespace descartes_utilities
+{
+  /**
+   * @brief Converts a sequence of Descartes joint trajectory points to ROS trajectory points.
+   *        Copies timing if specified, and sets vel/acc/effort fields to zeros.
+   * @param model Descartes robot model associated with the Descartes joint trajectory
+   * @param joint_traj Sequence of 'joint trajectory points' as returned by Dense/Sparse planner 
+   * @param default_joint_vel If a point, does not have timing specified, this value (in rads/s)
+   *                          is used to calculate a 'default' time. Must be > 0 & less than 100.
+   * @param out Buffer in which to store the resulting ROS trajectory. Only overwritten on success.
+   * @return True if the conversion succeeded. False otherwise.
+   */
+  bool
+  toRosJointPoints(const descartes_core::RobotModel& model,
+                   const std::vector<descartes_core::TrajectoryPtPtr>& joint_traj,
+                   double default_joint_vel,
+                   std::vector<trajectory_msgs::JointTrajectoryPoint>& out);
+}
+
+#endif
+
--- a/descartes_utilities/package.xml
+++ b/descartes_utilities/package.xml
@@ -1,0 +1,25 @@
+<?xml version="1.0"?>
+<package>
+  <name>descartes_utilities</name>
+  <version>0.0.1</version>
+  <description>
+    This package contains helper routines for working with the Descartes motion planning library
+    that ease practical use, but do fit cleanly into the core library. This includes conversions
+    to ROS trajectories, and similar operations.
+  </description>
+
+  <maintainer email="Jmeyer@swri.org">Jonathan Meyer</maintainer>
+
+  <license>Apache 2.0</license>
+
+  <buildtool_depend>catkin</buildtool_depend>
+  
+  <build_depend>descartes_core</build_depend>
+  <build_depend>descartes_trajectory</build_depend>
+  <build_depend>trajectory_msgs</build_depend>
+  
+  <run_depend>descartes_core</run_depend>
+  <run_depend>descartes_trajectory</run_depend>
+  <run_depend>trajectory_msgs</run_depend>
+
+</package>
--- a/descartes_utilities/src/ros_conversions.cpp
+++ b/descartes_utilities/src/ros_conversions.cpp
@@ -1,0 +1,121 @@
+/*
+ * Software License Agreement (Apache License)
+ *
+ * Copyright (c) 2016, Southwest Research Institute
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/*
+ *  ros_conversions.cpp
+ *
+ *  Created on: Feb 28, 2016
+ *  Author: Jonathan Meyer
+ */
+
+#include "descartes_utilities/ros_conversions.h"
+#include <algorithm>
+#include <console_bridge/console.h>
+
+/**
+ * @brief Given two sets of joint values representing two robot joint poses, this function computes the
+ *        minimum amount of time required to interpolate between the points assuming that each joint can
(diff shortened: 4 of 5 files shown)


descartes_utilities/include/descartes_utilities/ros_conversions.h
```diff
@@ -0,0 +1,50 @@
...
+#ifndef DESCARTES_ROS_CONVERSIONS_H
+#define DESCARTES_ROS_CONVERSIONS_H
+
+#include <trajectory_msgs/JointTrajectory.h>
+#include <descartes_core/trajectory_pt.h>
+
+namespace descartes_utilities
+{
+  /**
+   * @brief Converts a sequence of Descartes joint trajectory points to ROS trajectory points.
```
jrgnicho, line 34: I think it should be mentioned that the velocity, acceleration and effort arrays are filled with zeros.

shaun-edwards, line 34: Is this the correct behavior or should the arrays be left empty?

Jmeyer1292, line 34: Well, whether it's correct or not, some of our drivers including the [UR](https://github.com/ros-industrial/universal_robot/blob/indigo-devel/ur_driver/src/ur_driver/driver.py#L718) won't work without velocities being set to the same length as positions.

shaun-edwards, line 34: Is it possible to set the velocities based on timing, similar to how MoveIt does it?

gavanderhoorn, line 34: We ran into issues with this when trying to execute trajectories - converted using any of the 'boilerplate' versions of the descartes-to-moveit code - on robots other than the UR.

The code in the [roscon_2015](https://github.com/ros-industrial-consortium/roscon_2015) repository generates trajectories in the same way, and it seems to only result in smooth motion due to the dynamics of the UR and the relatively short traj segments. If the timing of the individual segments is changed, we quickly experienced the UR performing a sort of _stop-motion_. I expect the same for drivers like `motoman_driver` which depend heavily on the values in the `velocity` and `acceleration` fields to be able to resample the incoming segments.

We sometimes use the `IterativeParabolicTimeParameterization` from MoveIt to fixup trajectories, but that's all in joint space.

descartes_utilities/src/ros_conversions.cpp
```diff
@@ -0,0 +1,107 @@
...
+ *
+ *  Created on: Feb 28, 2016
+ *  Author: Jonathan Meyer
+ */
+
+#include "descartes_utilities/ros_conversions.h"
+#include <algorithm>
+#include "console_bridge/console.h"
+
+static double minTime(const std::vector<double>& a, const std::vector<double>& b, double max_vel)
```
gavanderhoorn: Suggestion: add some comments to document this method's parameters.

gavanderhoorn: Suggestion 2: avoid single letter variable names.

descartes_utilities/src/ros_conversions.cpp
```diff
@@ -0,0 +1,107 @@
...
+/*
+ *  ros_conversions.cpp
+ *
+ *  Created on: Feb 28, 2016
+ *  Author: Jonathan Meyer
+ */
+
+#include "descartes_utilities/ros_conversions.h"
+#include <algorithm>
+#include "console_bridge/console.h"
```
gavanderhoorn: Minor, but `console_bridge` is not a local include, so should use `<>` (this is not really required, but a convention).

descartes_utilities/src/ros_conversions.cpp
```diff
@@ -0,0 +1,107 @@
...
+  return *std::max_element(diff.begin(), diff.end());
+}
+
+bool
+descartes_utilities::toRosJointPoints(const descartes_core::RobotModel& model,
+                                      const std::vector<descartes_core::TrajectoryPtPtr>& joint_traj,
+                                      double default_joint_vel,
+                                      std::vector<trajectory_msgs::JointTrajectoryPoint>& out)
+{
+  if (default_joint_vel <= 0.0)
```
gavanderhoorn, line 55: Should there be an upper bound / sanity check on `default_joint_vel` as well?

descartes_utilities/src/ros_conversions.cpp
```diff
@@ -0,0 +1,107 @@
...
+  ros_trajectory.reserve(joint_traj.size());
+
+  std::vector<double> joint_point;
+  std::vector<double> dummy;
+
+  for (std::size_t i = 0; i < joint_traj.size(); ++i)
+  {
+    if (!joint_traj[i])
+    {
+      logError("%s: Input trajectory contained null pointer at index %lu", __FUNCTION__, i);
```
gavanderhoorn: nit-picking, but `std::size_t` doesn't necessarily fit `%lu` (depending on platform).

descartes_utilities/src/ros_conversions.cpp
Jmeyer1292: +1
I'll add some documentation, but I meant this function as a sort of private, implementation function.

descartes_utilities/src/ros_conversions.cpp
Jmeyer1292: Sure +1

descartes_utilities/src/ros_conversions.cpp
Jmeyer1292, line 55: What would it be?

What would be nice is some way to extract joint velocity limits from RobotModel.

gavanderhoorn, line 55: > What would it be?

yes, indeed. The problem is that you run the risk of making a _640k should be enough for everybody_-type of assumption.

However: a `double` has an insanely large range. It shouldn't be impossible to come up with something that is clearly out-of-bounds for this particular application.

descartes_utilities/src/ros_conversions.cpp
Jmeyer1292: +1
I'll switch it to `%zu`

gavanderhoorn: Yeah, so I had to fix this somewhere myself recently, and the issue is that `%zu` apparently is C99, and not necessarily portable across all C++ compilers (if they support it at all). Afaict (but I'd like to be wrong), the only really portable approach is to just cast to something that _does_ fit `%lu` (or whatever you want to use).

Alternatively, using C++ stream operators also works.

descartes_utilities/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+  descartes_core
+  descartes_trajectory
+  trajectory_msgs
+)
+
+catkin_package(
+ INCLUDE_DIRS include
+ LIBRARIES descartes_utilities
+ CATKIN_DEPENDS descartes_core descartes_trajectory trajectory_msgs
+)
```
gavanderhoorn, line 14: indent should be 2 spaces.

gavanderhoorn, line 14: it's recommended to sort `CATKIN_DEPENDS` items alphabetically.

descartes_utilities/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+
+include_directories(include)
+include_directories(
+  ${catkin_INCLUDE_DIRS}
+)
```
gavanderhoorn, line 19: just use single `include_directories(..)`?

descartes_utilities/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+
+add_library(${PROJECT_NAME}
+  src/ros_conversions.cpp
+)
+
+add_dependencies(${PROJECT_NAME} 
+  ${${PROJECT_NAME}_EXPORTED_TARGETS} 
+  ${catkin_EXPORTED_TARGETS}
+)
```
gavanderhoorn, line 27: If this package doesn't generate any messages / services of its own, then I think `${${PROJECT_NAME}_EXPORTED_TARGETS}` isn't actually needed.

descartes_utilities/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+  ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+  LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+  RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
+)
+
+install(DIRECTORY include/${PROJECT_NAME}/
+  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
+  FILES_MATCHING PATTERN "*.h"
+  PATTERN ".svn" EXCLUDE
+)
```
gavanderhoorn, line 38: I know this is in the boilerplate `CMakeLists.txt` that is generated by `catkin_create_pkg`, but the `EXCLUDE` pattern isn't really needed (we don't use subversion in this repository).

descartes_utilities/package.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0"?>
+<package>
```
gavanderhoorn, line 2: Might be nice to start using [package format 2](http://www.ros.org/reps/rep-0140.html).

descartes_utilities/include/descartes_utilities/ros_conversions.h
shaun-edwards, line 34: This might be why the `ur_modern_driver` had poor motion performance.

`IterativeParabolicTimeParameterization` would work since the trajectory is in joint space by this point in the process.  It wouldn't obey the timing constraints embedded in the trajectory though.

Jmeyer1292, line 34: @shaun-edwards We can fit a constrained cubic spline to each joint, and come up with smooth (if possibly out of range) velocities/accels for the entire trajectory without modifying the timing. I played with the concept using Eigen and got something going. I know everyone will want a library, so I've been looking:

I looked at the [KDL](http://docs.ros.org/hydro/api/orocos_kdl/html/classKDL_1_1VelocityProfile__Spline.html) spline function you mentioned, but it appears to apply to just one 'segment' between two points.

There's a very full featured solution in [ecl](http://wiki.ros.org/ecl_geometry/Tutorials/Splines) that will probably work, but it seems like quite the dependency.

The spline_smoother from the old arm nav package also had similar routines but I don't know if there's a catkin version.

@gavanderhoorn Are you aware of a good fit?

Jmeyer1292, line 34: Regardless, I think this is a bigger issue that needs to be addressed in another pull request. Short of adding velocity and acceleration constraints to the core descartes library, we could create a PR that implements one of the above filters and adds it as a "post-processing filter" to this here `descartes_utilities` package.

Can we accept something akin to this PR and log an issue for post-processing filters?

descartes_utilities/src/ros_conversions.cpp
Jmeyer1292: +1

descartes_utilities/CMakeLists.txt
Jmeyer1292, line 14: It is...

descartes_utilities/package.xml
Jmeyer1292, line 2: I'd rather leave it consistent with the rest of Descartes.

descartes_utilities/CMakeLists.txt
Jmeyer1292, line 38: +1

descartes_utilities/CMakeLists.txt
Jmeyer1292, line 27: +1

descartes_utilities/CMakeLists.txt
Jmeyer1292, line 19: +1

descartes_utilities/CMakeLists.txt
gavanderhoorn, line 14: You're right. My apologies.

shaun-edwards:
@Jmeyer1292, is this ready to merge (with an issue list perhaps)?

Jmeyer1292:
If you'll +1 it, I'm okay with it. 

I'll log an issue reflecting the need for velocity & acceleration parametrisation on the output of our trajectories. Anything else at the moment?

shaun-edwards:
+1 (after the issue is added).

descartes_utilities/include/descartes_utilities/ros_conversions.h
gavanderhoorn, line 34: @shaun-edwards wrote:

> `IterativeParabolicTimeParameterization` would work since the trajectory is in joint space by this point in the process.  It wouldn't obey the timing constraints embedded in the trajectory though.

Exactly. It's only a work-around really, not a solution. Ideally `descartes` would be able to do this itself, and then also based on Cartesian constraints.

gavanderhoorn, line 34: @Jmeyer1292 wrote:

> @gavanderhoorn Are you aware of a good fit?

No, I'm afraid not. I haven't spent too much time on this myself, just reporting what some colleagues of my ran in to.

PR #149, opened 2016-03-13
Removed Dynamic Allocation in default isValidMove()

Jmeyer1292:
As title says, the dynamic allocation in the isValidMove check was taking up like 15% of overall execution (with collision checking disabled). This allocation was unnecessary and has been replaced by a version that is logically equivalent without the use of dynamic memory.

### DIFF
--- a/descartes_moveit/src/moveit_state_adapter.cpp
+++ b/descartes_moveit/src/moveit_state_adapter.cpp
@@ -369,9 +369,6 @@ bool MoveitStateAdapter::isValidMove(const std::vector<double>& from_joint_pose,
                                      const std::vector<double>& to_joint_pose,
                                      double dt) const
 {
-  std::vector<double> max_joint_deltas;
-  max_joint_deltas.reserve(velocity_limits_.size());
-
   // Check for equal sized arrays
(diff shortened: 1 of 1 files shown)


descartes_moveit/src/moveit_state_adapter.cpp
```diff
@@ -379,18 +376,12 @@ bool MoveitStateAdapter::isValidMove(const std::vector<double>& from_joint_pose,
...
   }
 
-  // Build a vector of the maximum angle delta per joint 
-  for (std::vector<double>::const_iterator it = velocity_limits_.begin(); it != velocity_limits_.end(); ++it)
-  {
-    max_joint_deltas.push_back((*it) * dt);
-  }
-
-  for (std::vector<double>::size_type i = 0; i < from_joint_pose.size(); ++i)
+  // New version of isValidMove that doesn't rely on dynamic memory allocation
```
jrgnicho: I don't think this comment describing the fix needs to be permanently here

Jmeyer1292: +1

PR #150, opened 2016-03-14
Various Cleanups in Planning Graph

Jmeyer1292:
This PR helps assuage my OCD by doing a few minor, but important things to the core planning graph source file:
1. Moves several key INFO print statements to DEBUG statements. Descartes should now output no more than 3 lines on success. Warnings unchanged.
2. Replaces iterator types with `auto` keyword and created some high level typedefs. This makes the code shorter, easier to read, and should allow us to experiment with other data structures (perhaps a hash-map instead of a balanced tree?) more easily.
3. `std::list` is replaced with `std::vector` everywhere but the primary interfaces. There's no need for the properties of list, so vector gets us better memory usage, better cache coherency, etc... 

Unit tests still pass and I made sure to test with a Godel simulation.

### DIFF
--- a/descartes_planner/include/descartes_planner/planning_graph.h
+++ b/descartes_planner/include/descartes_planner/planning_graph.h
@@ -74,6 +74,7 @@ struct CartesianPointInformation
 };
 
 typedef std::map<descartes_core::TrajectoryPt::ID, CartesianPointInformation> CartesianMap;
+typedef std::map<descartes_core::TrajectoryPt::ID, descartes_trajectory::JointTrajectoryPt> JointMap;
 
 class PlanningGraph
 {
@@ -144,11 +145,11 @@ protected:
   //       and include an accessor to both formats
 
   // maintains the original (Cartesian) points list along with link information and associated joint trajectories per point
-  std::map<descartes_core::TrajectoryPt::ID, CartesianPointInformation> *cartesian_point_link_;
+  CartesianMap* cartesian_point_link_;
 
   // maintains a map of joint solutions with it's corresponding graph vertex_descriptor
   //   one or more of these will exist for each element in trajectory_point_map
-  std::map<descartes_core::TrajectoryPt::ID, descartes_trajectory::JointTrajectoryPt> joint_solutions_map_;
+  JointMap joint_solutions_map_;
 
   /** @brief simple function to iterate over all graph vertices to find ones that do not have an incoming edge */
   bool findStartVertices(std::list<JointGraph::vertex_descriptor> &start_points);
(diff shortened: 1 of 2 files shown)


descartes_planner/src/planning_graph.cpp
```diff
@@ -51,31 +51,10 @@ PlanningGraph::~PlanningGraph()
...
-  while(!done)
-  {
-    to_return[cart_id] = (*cartesian_point_link_)[cart_id];
-    cart_id = (*cartesian_point_link_)[cart_id].links_.id_previous;
-    done = (cart_id.is_nil());
-    ROS_DEBUG_STREAM("Next CID: " << cart_id);
-  }
-
-  return to_return;
+  if (!cartesian_point_link_) return CartesianMap();
```
jrgnicho, line 56: Not sure that this needs be enclosed within **if-else** statements as the result appears to be the same regardless of the pointer value.

Jmeyer1292, line 56: The if/else here is an attempt to guard against a segfault if this function is called prior to `insertGraph` and `cartesian_point_link_` is still null.

descartes_planner/include/descartes_planner/planning_graph.h
```diff
@@ -144,17 +146,17 @@ class PlanningGraph
   //       and include an accessor to both formats
 
   // maintains the original (Cartesian) points list along with link information and associated joint trajectories per point
-  std::map<descartes_core::TrajectoryPt::ID, CartesianPointInformation> *cartesian_point_link_;
+  CartesianMap* cartesian_point_link_;
```
jrgnicho, line 149: This might be a good opportunity to remove the pointer syntax from this member unless you think it needs to stay a pointer.

Jmeyer1292, line 149: Well, I intend to re-factor a lot of things going forward. That use of heap allocation is one. I can add it to this one, but there will be a continuous stream of these PR's fixing small things.

descartes_planner/src/planning_graph.cpp
jrgnicho, line 56: +1

descartes_planner/include/descartes_planner/planning_graph.h
jrgnicho, line 149: OK, then you may disregard my comment on this

jrgnicho, line 149: Looks good to me, +1 for merge

Jmeyer1292:
Thanks for reviewing this.

Issue #151, opened 2016-03-18
Descartes Planning Spends a Lot of Time Calculating Edges


Jmeyer1292:
It appears that a large portion of the time spent building a planning graph is actually spent looking up joint positions from the joint solutions std::map. I believe we can refactor the `insertGraph` method to reduce this by passing the joint vectors generated during IK expansion directly to the edge builders.

shaun-edwards:
@Jmeyer1292, can you provide the call graph and the details of your test case here.  It would be ideal if you could provide the test code/procedure as well.

Jmeyer1292:
@shaun-edwards 

Here's a pretty typical example as produced by Valgrinds' callgrind tool. Note the right-most branch. Far more time is spent building edges than doing IK (at least when using some closed-form equivalent). The `linearWeight` function is doing the actual math but it is less than half of the cost of it's parent. Compiling under `RelWithDebInfo` indicates that this time is spent (mostly) searching the `std::map`.
![graph](https://cloud.githubusercontent.com/assets/4960180/13925770/a2d0c9e4-ef57-11e5-9249-9dfddea71d80.jpg)

Jmeyer1292:
As far as test code/procedure, I've built a little package [here](https://github.com/Jmeyer1292/descartes/tree/benchmarks). It essentially uses the Roscon-2015 demo with some tweaks. It needs some work before it's ready to put into our main-line branch.

Jmeyer1292:
#154 at least partially addresses this issue. I'm going to close this issue for the sake of clarity. Further performance questions will be continued in new threads.

PR #154, opened 2016-03-21
Edge Building Refactor

Jmeyer1292:
Related to #151, this PR refactors the graph building process by generating all of the joint solutions and then putting off adding them to the tree until after all the edges have been calculated. This prevents us from doing a lot of unnecessary tree searching.

To accomplish this, I needed to create a `calculateEdges` function that took a sequence of joint positions, not trajectory IDs. This function is used in many other places in the code, however, and I didn't want to touch those. Someday soon I'll get around to looking carefully at the add/remove/modify functions.

### DIFF
--- a/descartes_planner/src/planning_graph.cpp
+++ b/descartes_planner/src/planning_graph.cpp
@@ -71,10 +71,9 @@ bool PlanningGraph::insertGraph(const std::vector<TrajectoryPtPtr> *points)
     ROS_ERROR_STREAM("points == null. Cannot initialize graph with null list.");
     return false;
   }
-  if (points->size() == 0)
+  if (points->size() < 2)
   {
-    // one or both have 0 elements
-    ROS_ERROR_STREAM("points.size == 0. Cannot initialize graph with 0 elements.");
+    ROS_ERROR_STREAM(__FUNCTION__ << ": must provide at least 2 input trajectory points.");
     return false;
   }
 
@@ -116,7 +115,6 @@ bool PlanningGraph::insertGraph(const std::vector<TrajectoryPtPtr> *points)
   std::vector<std::vector<JointTrajectoryPt>> poses;
   if (!calculateJointSolutions(*points, poses))
   {
-    // failed to get joint trajectories
     ROS_ERROR_STREAM("unable to calculate joint trajectories for input points");
     return false;
   }
@@ -124,7 +122,6 @@ bool PlanningGraph::insertGraph(const std::vector<TrajectoryPtPtr> *points)
   std::vector<JointEdge> edges;
   if (!calculateAllEdgeWeights(poses, edges))
   {
-    // failed to get edge weights
     ROS_ERROR_STREAM("unable to calculate edge weight of joint transitions for joint trajectories");
     return false;
   }
@@ -138,7 +135,6 @@ bool PlanningGraph::insertGraph(const std::vector<TrajectoryPtPtr> *points)
   // from list of joint trajectories (vertices) and edges (edges), construct the actual graph
   if (!populateGraphEdges(edges))
   {
-    // failed to create graph
     ROS_ERROR_STREAM("unable to populate graph from calculated edges");
     return false;
   }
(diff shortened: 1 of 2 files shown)


descartes_planner/src/planning_graph.cpp
```diff
@@ -113,26 +112,25 @@ bool PlanningGraph::insertGraph(const std::vector<TrajectoryPtPtr> *points)
...
     ROS_ERROR_STREAM("unable to calculate joint trajectories for input points");
     return false;
   }
 
-  if (!populateGraphVertices())
+  std::vector<JointEdge> edges;
+  if (!calculateAllEdgeWeights(poses, edges))
   {
-    ROS_ERROR_STREAM("unable to populate graph from input points");
+    // failed to get edge weights
```
jrgnicho: Having a commend and an error message that describe about the same feels redundant.

descartes_planner/src/planning_graph.cpp
```diff
@@ -763,95 +761,53 @@ void PlanningGraph::printGraph()
...
-    ROS_DEBUG_STREAM("CartID: " << tempID << " JointPoses count: " << joint_poses.size());
+    std::vector<std::vector<double>> joint_poses;
+    points[i]->getJointPoses(*robot_model_, joint_poses);
 
-    if (joint_poses.size() == 0)
+    if (joint_poses.empty())
     {
-      ROS_WARN_STREAM("no joint solution for this point... potential discontinuity in the graph");
+      ROS_ERROR_STREAM(__FUNCTION__ << ": IK failed for input trajectory point with ID = " << points[i]->getID());
+      return false;
```
jrgnicho, line 774: I'm surprised this wasn't returning **false** already.  It feels to me as that is the correct way to handle this when there are no joint solution; however were we deliberately ignoring this before?

descartes_planner/src/planning_graph.cpp
```diff
@@ -763,95 +761,53 @@ void PlanningGraph::printGraph()
...
+      ROS_ERROR_STREAM(__FUNCTION__ << ": unable to calculate any valid transitions between inputs " << (i-1) << " and " << i); 
+      return false;
     }
   }
 
   return !edges.empty();
 }
 
-bool PlanningGraph::calculateEdgeWeights(const std::vector<TrajectoryPt::ID> &start_joints,const std::vector<TrajectoryPt::ID> &end_joints, std::vector<JointEdge> &edge_results)
+bool PlanningGraph::calculateEdgeWeights(const std::vector<TrajectoryPt::ID> &start_joints,
```
jrgnicho, line 806: Since this method might potentially be deprecated we should log it as an issue so that it's not forgotten.

descartes_planner/src/planning_graph.cpp
Jmeyer1292, line 774: Yep, the process will continue often until the graph-search stage when you'll see a minimum cost equal to `std::numeric_limits<double>::max()`.

descartes_planner/src/planning_graph.cpp
Jmeyer1292, line 806: I'm not exactly sure what to do with this function until I get around to re-factoring the graph modification methods.

Jmeyer1292:
@jrgnicho I removed a few extra comments that were redundant in light of the debug statements they preceded.

jrgnicho:
The travis-ci check fails, not sure what that is about as I assumed you are able to build this.  That aside,  you can merge.

Jmeyer1292:
@jrgnicho So the Travis build works on the main ROS packages but fails with the package source set to "ros-shadow-fixed".

I'm not sure what to do about the error though. When this PR was first put up the tests passed. Removing comments clearly doesn't change anything meaningful so this appears to be an outside issue. Any thoughts?

```
+++rospack plugins --attrib=plugin nodelet
[rospack] Error: no such package nodelet
```

130s:
Sorry for the inconvenience. https://github.com/ros-industrial/industrial_ci/issues/26 is fixed now. So re-run Travis (you should be able to do so from https://travis-ci.org/ros-industrial-consortium/descartes/jobs/121059603) and see. If you still see the similar issue please report at ros-industrial/industrial_ci#26.

Jmeyer1292:
@130s Thank you.

@jrgnicho Since this is fixed, I'm merging.

Issue #148, opened 2016-03-13
Lack of Velocity / Acceleration Information in Descartes' Paths


Jmeyer1292:
Per discussion in #144. The output of Descartes planners is a joint path that obeys the timing constraints specified by the user. These timing constraints are used to perform some rough pruning of the trajectory graph, but are otherwise ignored.

In lieu of calculating velocity and acceleration values for each point as part of the planning problem, we could perform velocity parameterization as part of some post-processing filter. There are some questions to be answered still, such as how we will capture velocity/acceleration limits and what happens when a filter fails those limits, but I think this is the easiest path forward.

PR #157, opened 2016-03-29
Spline Interpolator Class & Utility Functions

Jmeyer1292:
This PR addresses #148. It creates a `SplineInterpolator` class that fits natural cubic splines of the form `a + bx + cx^2 + dx^3` to each pair of subsequent points.

New utility functions are now available in `parameterization.h`. These allow a user to build a set of splines from an input ROS trajectory, or simply parametrize a ROS trajectory from its `positions` and `time_from_start` fields.

The spline class also enables checking for maximum velocity and acceleration in the entire path.

I need to do some physical testing to make sure this produces desirable results, and I will update this PR when I have done that.

Furthermore, I'm curious for input on:
1. The interfaces in `parameterization.h`
2. Error handling. Currently all of the building is done in the constructor of `SplineInterpolator` but this could fail if the user provides invalid input. The utility functions check for these failure points, but anyone using the class directly may find issues. I added asserts to check for debug builds. I know there will be resistance here, but can anyone think of an elegant way to handle this without resorting to an `init` function? I'm hesistant to use exceptions because nothing else in Descartes does so.

### DIFF
--- a/descartes_utilities/CMakeLists.txt
+++ b/descartes_utilities/CMakeLists.txt
@@ -10,7 +10,10 @@ find_package(catkin REQUIRED COMPONENTS
 catkin_package(
   INCLUDE_DIRS include
   LIBRARIES descartes_utilities
-  CATKIN_DEPENDS descartes_core descartes_trajectory trajectory_msgs
+  CATKIN_DEPENDS 
+    descartes_core 
+    descartes_trajectory 
+    trajectory_msgs
 )
 
 include_directories(
@@ -20,7 +23,10 @@ include_directories(
 
 add_library(${PROJECT_NAME}
   src/ros_conversions.cpp
+  src/spline_interpolator.cpp
+  src/parameterization.cpp
 )
+target_link_libraries(${PROJECT_NAME} ${catkin_LIBRARIES})
 
 add_dependencies(${PROJECT_NAME} 
   ${catkin_EXPORTED_TARGETS}
@@ -36,3 +42,12 @@ install(DIRECTORY include/${PROJECT_NAME}/
   DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
   FILES_MATCHING PATTERN "*.h"
 )
+
+if(CATKIN_ENABLE_TESTING)
+  set(UTEST_SRC_FILES 
+      test/utest.cpp
+      test/parameterization_tests.cpp
+  )
+  catkin_add_gtest(${PROJECT_NAME}_utest ${UTEST_SRC_FILES})
+  target_link_libraries(${PROJECT_NAME}_utest ${PROJECT_NAME})
+endif()
--- a/descartes_utilities/include/descartes_utilities/parameterization.h
+++ b/descartes_utilities/include/descartes_utilities/parameterization.h
@@ -48,11 +48,38 @@ bool toCubicSplines(const std::vector<trajectory_msgs::JointTrajectoryPoint>& tr
  * Calculates velocity and acceleration values for a joint trajectory based on natural cubic splines
  * fit to the input trajectory. This function assumes that the 'positions' field of the trajectory
  * has been filled out for each point and that the degrees of freedom does not change.
- * @param  traj The input trajectory on which to act; assumed to have 'positions' field filled out.
- *              This function will populate 'velocities' and 'accelerations'.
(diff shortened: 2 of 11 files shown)


gavanderhoorn:
Just an FYI: `ros_control` has a potentially re-usable implementation of a quintic spline interpolator [here](https://github.com/ros-controls/ros_controllers/blob/95667092175a535ba8e14ef99f9f9301c5e61ebf/joint_trajectory_controller/include/trajectory_interface/quintic_spline_segment.h).

Might be worthwhile to investigate (lower maintenance, tested implementation, etc, etc).

There is also code in [ecl_geometry](http://wiki.ros.org/ecl_geometry) (I think you referred to that in an earlier PR). It would be a dependency, but that might be worth it. See [ecl_geometry/Tutorials/Splines](http://wiki.ros.org/ecl_geometry/Tutorials/Splines) fi.

And just to be clear: I'm not criticising your work, just wondering whether we should try to avoid implementing yet-another spline interpolator over (ROS) joint trajectories. It would not only mean less code to maintain and debug, but would also allow us to re-use behaviour in a way, as the resulting trajectories will be close / comparable to those coming out of `ros_control` fi.

Jmeyer1292:
@gavanderhoorn Don't worry, I completely understand :smile: . I drew a lot of inspiration and even an algorithm from the `ecl` package you mentioned. I, personally, consider that to be a really big dependency. I'd be interested in hearing the opinions of others @jrgnicho  @shaun-edwards.

The `ros_control` package you point toward is interesting, but I don't think that particular file provides the functionality we want. It requires the user to specify velocity/acceleration/position at the start/end of each segment. If you specify only position and time, as we have in Descartes atm, then (I think) it only does linear interpolation.

Jmeyer1292:
Side note, does anyone know what's wrong with the `shadow` build that's failing:

```
0.000003 * /opt/ros/indigo/share/doc/liborocos-kdl

+++rospack plugins --attrib=plugin nodelet

[rospack] Error: no such package nodelet

++++error
```

130s:
> Side note, does anyone know what's wrong with the shadow build that's failing:
> 
> ```
> +++rospack plugins --attrib=plugin nodelet
> [rospack] Error: no such package nodelet
> 
> ++++error
> ```

Work in progress to work around this new error. https://github.com/ros-industrial/industrial_ci/issues/26

130s:
Sorry for the inconvenience. https://github.com/ros-industrial/industrial_ci/issues/26 is fixed now. So re-run Travis (you should be able to do so from https://travis-ci.org/ros-industrial-consortium/descartes/jobs/119140362) and see. If you still see the similar issue please report at ros-industrial/industrial_ci#26.

Jmeyer1292:
I've just added:
1. Some extra checks on input trajectories to catch common errors that will cause the fitting process to fail
2. The ability to resample trajectories using the input splines.
3. Extra & improved error statements

I've also run this code on a real UR5 and the values seem reasonable.

hsd-dev:
May I know why was this PR not merged? 

Does this obey joint limits?

hsd-dev:
The SplineInterpolator works fine for me except the 1st point using the function: 
```cpp
descartes_utilities::setDerivatesFromSplines(traj)
```

There is a "swing" at the start after which the trajectory is executed as expected. After looking at the planned trajectory, the 1st point's acceleration does not start from 0:

figure_pos
figure_acc



```
header: 
  seq: 0
  stamp: 0.000000000
  frame_id: 
joint_names[]
points[]
  points[0]: 
    positions[]
      positions[0]: 1.56273
      positions[1]: 0.0724486
      positions[2]: 1.34394
      positions[3]: 0.0285536
      positions[4]: 0.299376
      positions[5]: -0.811875
    velocities[]
      velocities[0]: 0
      velocities[1]: 0
      velocities[2]: 0
      velocities[3]: 0
      velocities[4]: 5.55112e-17
      velocities[5]: 0
    accelerations[]
      accelerations[0]: 2.07194
      accelerations[1]: -6.03849
      accelerations[2]: -2.29
      accelerations[3]: -0.171915
      accelerations[4]: 18.0422
      accelerations[5]: -1.31063
    effort[]
      effort[0]: 0
      effort[1]: 0
      effort[2]: 0
      effort[3]: 0
      effort[4]: 0
      effort[5]: 0
    time_from_start: 1.000000000
```
If the accelerations are not calculated and set to 0, then the behaviour is more as expected.

Issue #124, opened 2015-05-12
descartes with ikfast


filipmrc:
I've been struggling with getting descartes to work with IKfast. The getAllIK function always returns 1 solution which is equal to the start state of the robot. 

I think I have narrowed down the problem to the way  the IKfast moveit plugin works, when moveit::core::RobotState calls setFromIK, which then passes it on to ikfast_moveit_plugin::IKFastMoveitPlugin::SearchPositionIK which always picks the first solution if it is valid. Now I've checked and the solver always finds the same set of solutions , that made me think that always the same seed_state_ is generated, but that isn't the case. Could anyone help me out?

shaun-edwards:
@jrgnicho, can you respond.

jrgnicho:
@filipmrc , the behavior you described is consistent with the way that ikfast works.  The seed is ignored by ikfast so regardless of what you pass it will always return the first valid solution.  The ikfast implementation subclasses the kinematics::KinematicsBase interface in moveit, but unfortunately this interface doesn't expose all of the ik solutions even if it finds them (see [here](https://github.com/ros-planning/moveit_core/pull/231) for more on this issue).  For now, all I can suggest to use this [workaround](https://github.com/ros-industrial-consortium/godel/tree/hydro-devel/motoman_sia20d_descartes) for wrapping the ikfast into a descartes_core::RobotModel.  Alternatively, you could use the kdl solver which is a lot slower but it'll return multiple solutions when you call getAllIK().  
In regards to your first remark, are you referring to the **seed** as the start state of the robot?

filipmrc:
First of all thank you for your quick reply.

If I understood the code correctly, seeds are random joint states generated when descartes_core::RobotModel is initialized. They are then used as starting points for a solution by the solver. I am currently attempting the workaround you suggested so perhaps I will have more success with that.

jrgnicho:
Yes, the MoveitStateAdapter (a RobotModel implementation) behaves as you described.  However, it is meant to be used with the KDL kinematics plugin since the kdl numerical approach takes a seed in order to localize the search for a solution.

filipmrc:
There is just one thing that puzzles me. In my attempts to test descartes with ikfast I realized that all the returned solutions were equal to the current state of the robot simulation I started with moveit demo.launch . 

More specifically , when GetAllIK calls GetIK , in moveit_state_adapter.cpp:

[...]

PR #160, opened 2016-04-21
IKFast Moveit State Adapter

Jmeyer1292:
See #124 for all the recent hoopla. 

This PR introduces a new class: `IkFastMoveitStateAdapter` that provides a specialization of `getAllIK()` to call the new method that @jrgnicho got put into both moveit's kinematics plugin and the MoveIt Ikfast adapter (with some work remaining to be done). 

I also took the opportunity to do a bit of cleanup in the primary MoveitStateAdapter. The biggest change is that I removed the non-default constructor. It duplicated all of the code in the initialize function and added more complexity than was necessary.

Note that on Indigo this will not be useful until [this](https://github.com/ros-planning/moveit_ikfast/pull/57) PR has been approved. It should work fine on Jade. We'll probably have to rebuild most of the existing IKFast plugins if that hasn't been done already.

I have tested on my own motoman SIA20D model. I'll try and build/recompile a few extra models.

### DIFF
--- a/descartes_moveit/src/ikfast_moveit_state_adapter.cpp
+++ b/descartes_moveit/src/ikfast_moveit_state_adapter.cpp
@@ -1,0 +1,52 @@
+/*
+ * Software License Agreement (Apache License)
+ *
+ * Copyright (c) 2016, Jonathan Meyer
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "descartes_moveit/ikfast_moveit_state_adapter.h"
+#include "eigen_conversions/eigen_msg.h"
+
+bool descartes_moveit::IkFastMoveitStateAdapter::getAllIK(const Eigen::Affine3d &pose, 
+  std::vector<std::vector<double> > &joint_poses) const
+{
+  joint_poses.clear();
+  const auto& solver = joint_group_->getSolverInstance();
+
+  // Transform input pose
+  Eigen::Affine3d tool_pose = world_to_root_.frame * pose;
+
+  // convert to geometry_msgs ...
+  geometry_msgs::Pose geometry_pose;
+  tf::poseEigenToMsg(tool_pose, geometry_pose);
+  std::vector<geometry_msgs::Pose> poses = {geometry_pose};
+
+  std::vector<double> dummy_seed (getDOF(), 0.0);
+  std::vector<std::vector<double>> joint_results;
+  kinematics::KinematicsResult result;
+  kinematics::KinematicsQueryOptions options; // defaults are reasonable as of Indigo
+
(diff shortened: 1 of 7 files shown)


Jmeyer1292:
I overlooked some issues having to do with differences in the base/tool frames of an arbitrary MoveGroup and the kinematics solved by IkFast.

I'll update as I fix them.

Jmeyer1292:
@shaun-edwards I believe I have resolved the issues I was having.

I have added special parameters to this specialization that indicate the base & tool frame of the underlying ikfast implementation. These values default to `base_link` and `tool0`. Other functions have been overloaded to take these transformations into account.

Please review.

descartes_moveit/src/ikfast_moveit_state_adapter.cpp
```diff
@@ -0,0 +1,155 @@
...
+#include "descartes_moveit/ikfast_moveit_state_adapter.h"
+
+#include <cassert>
+#include <eigen_conversions/eigen_msg.h>
+#include <ros/node_handle.h>
+
+const static std::string default_base_frame = "base_link";
+const static std::string default_tool_frame = "tool0";
+
+// Compute the 'joint distance' between two poses
```
BrettHemes-3M, line 28: Could you elaborate on why you chose the 1-norm here?

BrettHemes-3M:
+1
I ran this today (along with with ros-planning/moveit_ikfast#57) with no issues.

descartes_moveit/src/ikfast_moveit_state_adapter.cpp
shaun-edwards, line 28: @BrettHemes vs. an L2-norm?

BrettHemes-3M, line 28: Yeah... I would have defaulted to L2 but as I typed the comment I became less sure of my choice and now I am just curious.

shaun-edwards, line 28: L2 is probably mathematically more correct.  However, an L1 is more efficient (no square root or multiplication)  If all that is considered from the distance function is a value comparison (> or <) and not a true distance difference, then the resulting comparison is the same whether an L2 or L1 norm is used.

shaun-edwards:
I'll let @Jmeyer1292 do the merge, but I wanted to thank @BrettHemes  for reviewing.

descartes_moveit/src/ikfast_moveit_state_adapter.cpp
Jmeyer1292, line 28: @BrettHemes @shaun-edwards I'm rather embarrassed to admit there was no great thought put into this choice of norm for computing the distance between poses. I will at least say that, for better or worse, the L1 matches the cost function used to compute edge costs when solving our graphs.

I can change this if you think it's more correct.

shaun-edwards, line 28: The L1 is correct, assuming the `distance` function is not accessible outside the class.  If it is, then distance should mean distance (L2 norm).

Efficiency is important in this class, so use of the L1 even though it isn't exactly correct is better.  Of course this assumes I'm right about the relative complexity of L1 vs L2 (given our discussion at lunch, I can't be sure).

Jmeyer1292:
Thanks @BrettHemes for reviewing. Please continue to let us know as you have needs & issues.

For everyone, please note that this PR requires [ ros-planning/moveit_ikfast#57](https://github.com/ros-planning/moveit_ikfast/pull/57) to be effective.

yishinli:
@Jmeyer1292 

How to set ikfast_tool_frame if its name is not "tool0"?
Ans: In .launch file, define the param for "/" node.

```
<launch>
  <param name="ikfast_tool_frame" value="tool"/>
</launch>
```

PR #161, opened 2016-05-03
moveit_state_adapter: fix velocity/acceleration setting for robots with passive joints

achim-k:
Hi,
my robot has 6 active and 5 passive joints. When executing the `descartes_tutorial` example, I got an error caused in this [line](https://github.com/ros-planning/moveit_core/blob/7a8fc418268173bf25b6a979784068f3d3f241e1/robot_state/include/moveit/robot_state/robot_state.h#L244). 
The problem is, that `setVariableVelocities(const std::vector<double> &velocity)` expects the size of the `velocity` vector to >= the size of joints of the robot_model (which includes the passive joints apparently). Same yields for `setVariableAccelerations(...)`.

This PR fixes this issue by setting the velocities/accelerations for only the active joints, using `void setVariableVelocities(const std::map<std::string, double> &variable_map)`

### DIFF
--- a/descartes_moveit/src/moveit_state_adapter.cpp
+++ b/descartes_moveit/src/moveit_state_adapter.cpp
@@ -321,8 +321,15 @@ bool MoveitStateAdapter::isValid(const std::vector<double> &joint_pose) const
     robot_state_->setJointGroupPositions(group_name_, joint_pose);
     //TODO: At some point velocities and accelerations should be set for the group as
     //well.
-    robot_state_->setVariableVelocities(std::vector<double>(joint_pose.size(), 0.));
-    robot_state_->setVariableAccelerations(std::vector<double>(joint_pose.size(), 0.));
+
+    const std::vector<std::string> active_joint_names = robot_state_->getJointModelGroup(group_name_)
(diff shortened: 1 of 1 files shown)


Jmeyer1292:
Hey! I appreciate the PR. There are a bunch of people who have issues surrounding this function.

My only concern is performance related: The `isValid` function is probably the most called function in all of Descartes (100,000's of calls). I want to try your PR under some benchmarks to see how it affects things.

Correctness is more important than speed here, but the bounds checking on velocity/acceleration is completely unused at the moment. Furthermore, I'm pretty sure that most kinematics plugins already filter by joint limits but of course there's no guarantee of that in our interface.

If you remove the need to set/check velocities/acceleration limits, can you think of a more efficient implementation of joint limit checking?

achim-k:
I admittedly know too little about Descartes/Path planning to give a valid answer here :smile: But I will have a look at it

Jmeyer1292:
@achim-k I decided to just remove the code that was causing you issues and replace it with an alternative. If you time and interest, could you test out #164 to verify that it works with the model you're using?

Jmeyer1292:
Closed; Issue addressed by #164.

Issue #130, opened 2015-07-09
Support of manipulators with more than 1 groups of joints


vkhamianok:
The "descartes" library fails in the function:

bool MoveitStateAdapter::isValid(const std::vector<double> &joint_pose) const

 with message:

```
/opt/ros/indigo/include/moveit/robot_state/robot_state.h:244: void moveit::core::RobotState::setVariableVelocities(const std::vector<double>&): Assertion `robot_model_->getVariableCount() <= velocity.size()' failed.
```

when I'm trying to calculate a Forward Kinematics using the function:

```
  virtual bool getFK(const std::vector<double> &joint_pose, Eigen::Affine3d &pose) const = 0;
```

for a manipulator with 2 (or more) groups of joints.

The problem is that active joints in a JointModelGroup named as "group_name_" is just a subset of all manipulators' joints. So number of active joints in JointModelGroup is less than VariableCount of a robot_state_ of such manipulator. So either this check fails:

```
  if (robot_state_->getJointModelGroup(group_name_)->getActiveJointModels().size() ==
      joint_pose.size())
```

if you pass a std::vector with positions for every joint_pose for a manipulator. Or:

```
    robot_state_->setVariableVelocities(std::vector<double>(joint_pose.size(), 0.));
```

fails (with the message shown in the beginning) if you pass a std::vector with positions just for a group. 

Seems that we need to change:

```
    robot_state_->setVariableVelocities(std::vector<double>(joint_pose.size(), 0.));
    robot_state_->setVariableAccelerations(std::vector<double>(joint_pose.size(), 0.));
    if (robot_state_->satisfiesBounds())
```

to:

```
    robot_state_->setVariableVelocities(robot_state_->getJointModelGroup(group_name_)->getActiveJointModelNames(), std::vector<double>(joint_pose.size(), 0.));
    robot_state_->setVariableAccelerations(robot_state_->getJointModelGroup(group_name_)->getActiveJointModelNames(), std::vector<double>(joint_pose.size(), 0.));
    if (robot_state_->satisfiesBounds(robot_state_->getJointModelGroup(group_name_)))
```

But I'm not very familiar with "descartes" (and whole ROS) code and not sure if it the appropriate changes.

Jmeyer1292:
Yeah, this has been an issue that has come up a few times but I haven't gotten around to fixing it yet.

The temporary fix is to do the following:

Goto line 324 and 325 of moveit_state_adapter.cpp in descartes_moveit/

[...]

PR #164, opened 2016-05-23
MoveIt-State-Adapter RobotState handling improvements

Jmeyer1292:
Related to #161, #130.

This PR:
1. Removes checks on accelerations/velocities that do not currently serve a purpose
2. Refactors the `isValid` function to hopefully work with a wider range of MoveIt robots (e.g. those that have passive joints or extra joints not in the move-group being planned for).
3. Adds a `setState` function to the two moveit robot models that allows a user to set the initial state for the planner. This enables a user to plan with robots that have additional joints before the desired planning group. You could, for example, set the position of a rail, get the proper transforms, and then plan for the 6 DOF robot ontop.

No performance regressions (in fact a small improvement), and the basic tests and robot models I have work.

I'll come up with some extras to test.

### DIFF
--- a/descartes_moveit/include/descartes_moveit/moveit_state_adapter.h
+++ b/descartes_moveit/include/descartes_moveit/moveit_state_adapter.h
@@ -83,6 +83,17 @@ public:
     return robot_state_;
   }
 
+  /**
+   * @brief Copies the internal state of 'state' into this model. Useful for initializing the
+   *        value of joints that are not part of the active move group. Should be called after
+   *        'initialize()'.
+   */
+  void setState(const moveit::core::RobotState& state)
+  {
+    assert(static_cast<bool>(robot_state_));
+    *robot_state_ = state;
+  }
+
 protected:
   /**
    * Gets IK solution (assumes robot state is pre-seeded)
--- a/descartes_moveit/include/descartes_moveit/moveit_state_adapter.h
+++ b/descartes_moveit/include/descartes_moveit/moveit_state_adapter.h
@@ -88,11 +88,7 @@ public:
    *        value of joints that are not part of the active move group. Should be called after
    *        'initialize()'.
    */
-  void setState(const moveit::core::RobotState& state)
-  {
-    assert(static_cast<bool>(robot_state_));
-    *robot_state_ = state;
-  }
+  void setState(const moveit::core::RobotState& state);
 
 protected:
   /**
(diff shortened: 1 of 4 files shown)


Jmeyer1292:
@jrgnicho please have a look. This one is straightforward, I think.

descartes_moveit/include/descartes_moveit/moveit_state_adapter.h
```diff
@@ -83,6 +83,17 @@ class MoveitStateAdapter : public descartes_core::RobotModel
...
   }
 
+  /**
+   * @brief Copies the internal state of 'state' into this model. Useful for initializing the
+   *        value of joints that are not part of the active move group. Should be called after
+   *        'initialize()'.
+   */
+  void setState(const moveit::core::RobotState& state)
+  {
+    assert(static_cast<bool>(robot_state_));
```
jrgnicho: Would it not be better to use ROS_ASSERT_MSG in order to provide a description of the type of error?

Jmeyer1292:
@jrgnicho I changed both asserts to ROS_ASSERT messages. Do you have any other concerns?

descartes_moveit/src/ikfast_moveit_state_adapter.cpp
```diff
@@ -126,7 +125,7 @@ bool descartes_moveit::IkFastMoveitStateAdapter::getFK(const std::vector<double>
 
 void descartes_moveit::IkFastMoveitStateAdapter::setState(const moveit::core::RobotState& state)
 {
-  assert(static_cast<bool>(robot_state_));
+  ROS_ASSERT_MSG(static_cast<bool>(robot_state_), "'robot_state_' member pointer is null. Have you called initialize()?");
   *robot_state_ = state;
```
jrgnicho: I may be contradicting myself here but rather that checking the nullity of the `robot_state_` pointer why not just call its **reset** method and use the [RobotState copy constructor](http://docs.ros.org/jade/api/moveit_core/html/classmoveit_1_1core_1_1RobotState.html#ae226721d6aff0342786c2acdc53a5d20) ?

Jmeyer1292: I want to force the user to call `initialize()` before they touch this function. If we made only that change, then the `robot_state_` is going to get blown away when they do call `initialize()`. 

We'd need to change that function to check if the state variable has been initialized ahead of the rest of the system. If you want to do that we can; I just saw this as the most expedient solution.

jrgnicho: As long as the appropriate usage of the interface is well documented I don't that is necessary.

Jmeyer1292: It's in the doc comment above the function, and all our tutorials indicate `robot->initialize()` is the first thing to do.

jrgnicho: You can merge

Jmeyer1292: @jrgnicho Thanks for reviewing

PR #174, opened 2016-06-18
Internal-Joint-Point

Jmeyer1292:
This PR replaces the use of `JointTrajectoryPt` inside the `planning_graph` with a `InternalJointPt` class that mimics the former's interface, but only includes components used by the planning graph.

This means that every vertex no longer stores additional upper/lower/discretization vectors or coordinate frame transforms.

On the ROSCON demo, with discretization of 0.05 radians, maximum memory consumption dropped 36%. With 0.1 radians, it dropped approximately 48%. Program execution time dropped by about 10%.

### DIFF
--- a/descartes_planner/include/descartes_planner/planning_graph.h
+++ b/descartes_planner/include/descartes_planner/planning_graph.h
@@ -54,6 +54,32 @@ struct CartesianPointRelationship
   descartes_core::TrajectoryPt::ID id_next;
 };
 
+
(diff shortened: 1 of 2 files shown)


jrgnicho:
Does this change not affect the high level Sparse and Dense planners?

Jmeyer1292:
It does not. The internal point mimics the interface of a `JointTrajectoryPoint` (ala Python duck-typing) and no modifications to the other components are required.

jrgnicho:
+1

PR #187, opened 2016-11-30
Make isInCollision(..) use an up-to-date PlanningScene instance when checking (for Godel)

gavanderhoorn:
As per subject. This appears to be the minimum amount of changes needed to implement the described functionality.

Note that in order to avoid service invocation overhead, callers of any planner functions are required to first call `robot_model::updateInternals()`.

This PR might not be mergeable as-is, but I wanted to allow for discussion here.


Jmeyer1292:
Merging this PR into a `kinetic-godel` branch for testing purposes.

gavanderhoorn:
Note that this might crash `move_group` sometimes, probably due to a bug in OctoMap (OctoMap/octomap#147).

PR #195, opened 2017-04-06
Fix for the ROS-Test in descartes_moveit

Jmeyer1292:
See the commit. Travis builds again!

### DIFF
--- a/descartes_moveit/package.xml
+++ b/descartes_moveit/package.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<package>
+<package format="2">
   <name>descartes_moveit</name>
   <version>0.0.1</version>
   <description>Moveit wrapper functions for descartes base types</description>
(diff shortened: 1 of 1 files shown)


descartes_moveit/package.xml
```diff
@@ -14,26 +14,18 @@
...
 
-  <build_depend>rosconsole_bridge</build_depend>
-  <build_depend>moveit_core</build_depend>
-  <build_depend>tf</build_depend>
-  <build_depend>cmake_modules</build_depend>
-  <build_depend>descartes_core</build_depend>
-  <build_depend>descartes_trajectory</build_depend>
-  <build_depend>moveit_ros_planning</build_depend>
-  <build_depend>pluginlib</build_depend>
-  <build_depend>eigen</build_depend>
```
AustinDeric: eigen wasn't previously in the run_depend, therefore it should be <build_depend> because it doesn't have a dependency at run-time.

http://www.ros.org/reps/rep-0140.html#build-depend-multiple

AustinDeric (changes requested):

Jmeyer1292:
Addressed comments by @AustinDeric. Will merge when travis completes.

PR #197, opened 2017-05-14
Trajectory Interpolator (Against Kinetic Devel)

Jmeyer1292:


descartes_utilities/package.xml
```diff
@@ -11,6 +11,7 @@
   <maintainer email="Jmeyer@swri.org">Jonathan Meyer</maintainer>
 
   <license>Apache 2.0</license>
+  <license>BSD</license>
```
jrgnicho, line 14: The license on the source files still says Apache 2.0

Jmeyer1292:
So I resurrected this PR after closing it a while ago after discussing with @Levi-Armstrong. It amused me that this is not too different from what MoveIt was doing and I wanted to keep the code around somewhere.

descartes_utilities/package.xml
Jmeyer1292: I lifted some code from the ecl library, which is licensed as BSD. I imagine I should include the BSD clause in that source file? I have no idea how this dual licensing stuff would work.

descartes_utilities/src/spline_interpolator.cpp
```diff
@@ -0,0 +1,232 @@
+/**
```
Levi-Armstrong, line 1: Is this the code that is BSD? If so, I think it needs to be labeled BSD similar to the others that are marked Apache.

descartes_utilities/src/spline_tests.cpp
```diff
@@ -0,0 +1,81 @@
...
+  //     j6 << point.positions[5] << "," << point.velocities[5] << "," << point.accelerations[5] << "\n";
+  //   }
+
+  //   std::cout << "Ji: " << points.front().positions[0] << " " << points.front().velocities[0] << " " << points.front().accelerations[0] << "\n";
+  //   std::cout << "Jf: " << points.back().positions[0] << " " << points.back().velocities[0] << " " << points.back().accelerations[0] << "\n";
+  // }
+  // else
+  // {
+  //   std::cout << "It did not work\n";
+  // }
```
jrgnicho, line 56: Commented code

PR #207, opened 2017-12-01
OMP Integration 2.0

Jmeyer1292:
Based partially on @BrettHemes work in #182. @BrettHemes please take a look.

Notable changes from your implementation:
1. I dropped the notion of cloning robot models and instead tried to make the collision checks thread safe. Now that I write this, I realize I need to look at the KDL-based functions and make them thread safe too.

2. I adopted your **success** flag for failing gracefully when building graph vertices. I confess that in a lot of the work I've done recently I have wanted to generate the graph even if a few points failed. I post process the graph down to something that can work. This is probably not the common use case though, so I've included your changes.

3. I parallelize at the outer layer of edge building, meaning that a single very expensive edge could totally circumvent any benefits. In past efforts I foolishly tried to micro optimize and made it hard on myself to implement your exact change. Would you please try this out and let me know the relative consequences w.r.t your work?

### DIFF
--- a/descartes_moveit/src/moveit_state_adapter.cpp
+++ b/descartes_moveit/src/moveit_state_adapter.cpp
@@ -245,8 +245,10 @@ bool MoveitStateAdapter::isInCollision(const std::vector<double>& joint_pose) co
   bool in_collision = false;
   if (check_collisions_)
   {
-    robot_state_->setJointGroupPositions(group_name_, joint_pose);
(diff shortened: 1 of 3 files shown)


BrettHemes-3M:
@Jmeyer1292 I finally got around to testing this (vs my omp fork) and it works great!  It seems to take about 75% of the planning time of the pre-refactored omp implementation for my use case; very nice!  :tada: 

Thanks for putting some much love and care into this package :smile:

PR #230, opened 2018-08-07
Adding collision checking capabilities to Descartes and preparing for changes to the IKFast templates

mlautman:
Adds collision checking capabilities to Descartes using the MoveIt! planning scene monitor. 
This is fundamentally different from the kinetic-godel version which is unable to take in a planning scene monitor which uses the planning scene spun up by the `demo.launch` script in moveit_config packages.
  
These changes are used by the Descartes MoveIt! capability that will soon be open sourced. It is unclear if these changes should live in a separate branch, fork, or if they should be merged into master. This PR is meant as a conversation starter. 

This PR is a result of a collaboration between [Carbon Robotics](http://www.carbon.ai/) and [PickNik Robotics](picknik.ai)

### DIFF
--- a/descartes_core/include/descartes_core/robot_model.h
+++ b/descartes_core/include/descartes_core/robot_model.h
@@ -23,6 +23,13 @@
 #include <moveit/kinematic_constraints/kinematic_constraint.h>
 #include "descartes_core/utils.h"
 
+
+namespace planning_scene_monitor
+{
+  MOVEIT_CLASS_FORWARD(PlanningSceneMonitor);
+  class PlanningSceneMonitor;
+}
+
 namespace descartes_core
 {
 DESCARTES_CLASS_FORWARD(RobotModel);
@@ -107,7 +114,8 @@ public:
    * a fixed location relative to the last link in 'group_name'.
    */
   virtual bool initialize(const std::string &robot_description, const std::string &group_name,
-                          const std::string &world_frame, const std::string &tcp_frame) = 0;
+                          const std::string &world_frame, const std::string &tcp_frame,
+                          planning_scene_monitor::PlanningSceneMonitorPtr psm = nullptr) = 0;
 
   /**
    * @brief Enables collision checks
--- a/descartes_moveit/include/descartes_moveit/ikfast_moveit_state_adapter.h
+++ b/descartes_moveit/include/descartes_moveit/ikfast_moveit_state_adapter.h
@@ -31,7 +31,8 @@ public:
   }
 
   virtual bool initialize(const std::string& robot_description, const std::string& group_name,
-                          const std::string& world_frame, const std::string& tcp_frame);
+                          const std::string& world_frame, const std::string& tcp_frame,
+                          planning_scene_monitor::PlanningSceneMonitorPtr psm = nullptr);
 
   virtual bool getAllIK(const Eigen::Affine3d& pose, std::vector<std::vector<double> >& joint_poses) const;
 
--- a/descartes_moveit/src/ikfast_moveit_state_adapter.cpp
+++ b/descartes_moveit/src/ikfast_moveit_state_adapter.cpp
@@ -17,6 +17,7 @@
  */
 
 #include "descartes_moveit/ikfast_moveit_state_adapter.h"
+#include <moveit/planning_scene_monitor/planning_scene_monitor.h>
 
 #include <eigen_conversions/eigen_msg.h>
 #include <ros/node_handle.h>
@@ -54,9 +55,10 @@ static size_t closestJointPose(const std::vector<double>& target, const std::vec
 bool descartes_moveit::IkFastMoveitStateAdapter::initialize(const std::string& robot_description,
                                                             const std::string& group_name,
                                                             const std::string& world_frame,
-                                                            const std::string& tcp_frame)
+                                                            const std::string& tcp_frame,
+                                                            planning_scene_monitor::PlanningSceneMonitorPtr psm)
 {
-  if (!MoveitStateAdapter::initialize(robot_description, group_name, world_frame, tcp_frame))
+  if (!MoveitStateAdapter::initialize(robot_description, group_name, world_frame, tcp_frame, psm))
   {
     return false;
   }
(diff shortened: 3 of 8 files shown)


descartes_moveit/src/ikfast_moveit_state_adapter.cpp
```diff
@@ -70,12 +72,9 @@ bool descartes_moveit::IkFastMoveitStateAdapter::getAllIK(const Eigen::Affine3d&
   joint_poses.clear();
   const auto& solver = joint_group_->getSolverInstance();
 
-  // Transform input pose
-  Eigen::Affine3d tool_pose = world_to_base_.frame_inv * pose * tool0_to_tip_.frame;
-
   // convert to geometry_msgs ...
   geometry_msgs::Pose geometry_pose;
-  tf::poseEigenToMsg(tool_pose, geometry_pose);
+  tf::poseEigenToMsg(pose, geometry_pose);
```
mlautman, line 77: Note: This fixes a double transform issue that will arrise once https://github.com/ros-planning/moveit/pull/1014 is merged.

descartes_moveit/include/descartes_moveit/ikfast_moveit_state_adapter.h
```diff
@@ -31,7 +31,8 @@ class IkFastMoveitStateAdapter : public descartes_moveit::MoveitStateAdapter
   }
 
   virtual bool initialize(const std::string& robot_description, const std::string& group_name,
-                          const std::string& world_frame, const std::string& tcp_frame);
```
mlautman, line 34: I'm pretty sure that we will no longer need the IkFastMoveitStateAdapter once the IKFast templates are updated https://github.com/ros-planning/moveit/pull/1014#issuecomment-410066739

mlautman:
@jrgnicho @Jmeyer1292

descartes_core/include/descartes_core/robot_model.h
```diff
@@ -107,7 +114,8 @@ class RobotModel
    * a fixed location relative to the last link in 'group_name'.
    */
   virtual bool initialize(const std::string &robot_description, const std::string &group_name,
-                          const std::string &world_frame, const std::string &tcp_frame) = 0;
+                          const std::string &world_frame, const std::string &tcp_frame,
+                          planning_scene_monitor::PlanningSceneMonitorPtr psm = nullptr) = 0;
```
Jmeyer1292, line 118: I don't know what others think, but I'm super dubious of putting so much MoveIt specific stuff in the constructor of what is supposed to be a generic kinematics class.

descartes_moveit/src/moveit_state_adapter.cpp
```diff
@@ -137,6 +140,15 @@ bool MoveitStateAdapter::initialize(robot_model::RobotModelConstPtr robot_model,
...
 
+  if(psm)
+  {
+    planning_scene_monitor_ = psm;
+  }
+  else
+  {
+    planning_scene_monitor_.reset(new planning_scene_monitor::PlanningSceneMonitor(robot_description_));
+  }
+  planning_scene_monitor_->startSceneMonitor();
```
Jmeyer1292, line 151: Does this imply that your planning  scene may change while in the middle of a request?

descartes_moveit/src/moveit_state_adapter.cpp
```diff
@@ -245,11 +260,17 @@ bool MoveitStateAdapter::isInCollision(const std::vector<double>& joint_pose) co
   bool in_collision = false;
   if (check_collisions_)
   {
-    moveit::core::RobotState state (robot_model_ptr_);
-    state.setToDefaultValues();
-    state.setJointGroupPositions(joint_group_, joint_pose);
-    in_collision = planning_scene_->isStateColliding(state, group_name_);
+    std::unique_ptr<planning_scene_monitor::LockedPlanningSceneRO> ls(new planning_scene_monitor::LockedPlanningSceneRO(planning_scene_monitor_));
```
Jmeyer1292, line 263: What is the cost of this? It seems like we're doing everything through a single robot state `robot_state_`. Is this thread safe?

descartes_moveit/src/moveit_state_adapter.cpp
```diff
@@ -245,11 +260,17 @@ bool MoveitStateAdapter::isInCollision(const std::vector<double>& joint_pose) co
...
+
+    // (*ls)->printKnownObjects(std::cout);
+
+    robot_state_->setJointGroupPositions(group_name_, joint_pose);
+    robot_state_->update();
+
+    // If either there is no locked planning scene or if the state is colliding return false
+    in_collision = !(*ls) || (*ls)->isStateColliding(*robot_state_, group_name_);
```
Jmeyer1292, line 271: What does it mean for there to be "no locked planning scene"?

descartes_moveit/src/moveit_state_adapter.cpp
```diff
@@ -295,7 +316,19 @@ bool MoveitStateAdapter::isValid(const std::vector<double>& joint_pose) const
     return false;
   }
 
-  return isInLimits(joint_pose) && !isInCollision(joint_pose);
+  // Satisfies joint positional bounds?
+  if (!isInLimits(joint_pose))
+  {
+    CONSOLE_BRIDGE_logError("Joint pose does not satisfy positional bounds");
```
Jmeyer1292, line 322: This really isn't an "error": These sort of poses get generated all the time by IK solvers. I use it to search for redundant, valid solutions.

Jmeyer1292 (commented):
I acknowledge that the interfaces in Descartes don't give a super clean way to implement things like collision checking or scene mutation.

My concern is that embedding a state listener ties us too closely to MoveIt and may mutate the state of the cell while in the middle of planning.

In the past I've added a non-const reference accessor to the planning scene in Moveit State Adapter. This way code outside of Descartes can make changes and plan "offline". Have you or would you consider such an approach for what you're trying to do?

I appreciate you making this PR and engaging with us.

descartes_core/include/descartes_core/robot_model.h
shaun-edwards: We went to great lengths to keep Descartes independent of MoveIt.  I believe that this is still true and at least some of the tests exercise Descartes independent of MoveIt, though it has been awhile.  

My suggested approach is to create a Descartes interface class and then have one of the implementations be the MoveIt planning scene OR why not just include the planning scene inside the `ikfast_moveit_state_adapter.h`?  While the robot model primarily performs kinematics, it's intent was to having collision checking embedded (otherwise it would have been called robot kinematics).

Jmeyer1292: What I've done for projects is create a function specific to `MoveItStateAdapter` and its children that looks like: `PlanningScene& getPlanningScene()`. This allows users to mutate the planning scene provided they know their robot is descended from the MoveIt adapter line of Descartes robot models.

What would that Descartes interface look like? Would it be our generic representation around the collision checking? E.g.
```c++

class CollisionInterface {
public:
  virtual bool isInCollision(const std::vector<double>& joints) const = 0;
}
```

And we would inject that into child classes?

descartes_moveit/src/moveit_state_adapter.cpp
mlautman: The planning scene is locked while planning

descartes_moveit/src/moveit_state_adapter.cpp
mlautman: The locked planning scene overwrites the bool operator to check that the internal planning_scene_monitor_ shared pointer is non-null and that it has a planning scene.

